PR #4, opened 2016-04-07
Retry on openApplication error (darwin); More error logging

gabriel:
It will retry open if it fails (after 3 seconds) for OS X... Adds some more logging for errors...

### DIFF
--- a/updater_osx.go
+++ b/updater_osx.go
@@ -6,6 +6,7 @@
 package updater
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"os/user"
@@ -68,9 +69,20 @@ func (u *Updater) checkPlatformSpecificUpdate(sourcePath string, destinationPath
 	return nil
 }
 
-func openApplication(applicationPath string) error {
-	_, err := exec.Command("/usr/bin/open", applicationPath).Output()
-	return err
+func (u *Updater) openApplication(applicationPath string) error {
+	tryOpen := func() error {
+		out, err := exec.Command("/usr/bin/open", applicationPath).Output()
+		if err != nil {
+			return fmt.Errorf("%s; %s", err, out)
+		}
+		return nil
+	}
+	if err := tryOpen(); err != nil {
+		u.log.Errorf("Open error, trying again in a few seconds; %s", err)
+		time.Sleep(3 * time.Second)
+		return tryOpen()
+	}
+	return nil
 }
 
 func (u *Updater) applyUpdate(localPath string) (err error) {
--- a/updater_osx.go
+++ b/updater_osx.go
@@ -77,10 +77,13 @@ func (u *Updater) openApplication(applicationPath string) error {
 		}
 		return nil
 	}
-	if err := tryOpen(); err != nil {
-		u.log.Errorf("Open error (trying again in a few seconds): %s", err)
-		time.Sleep(3 * time.Second)
-		return tryOpen()
+	for i := 0; i < 10; i++ {
+		err := tryOpen()
+		if err == nil {
+			break
+		}
+		u.log.Errorf("Open error (trying again in a second): %s", err)
+		time.Sleep(1 * time.Second)
 	}
 	return nil
 }
--- a/updater_osx.go
+++ b/updater_osx.go
@@ -71,9 +71,9 @@ func (u *Updater) checkPlatformSpecificUpdate(sourcePath string, destinationPath
 
 func (u *Updater) openApplication(applicationPath string) error {
 	tryOpen := func() error {
-		out, err := exec.Command("/usr/bin/open", applicationPath).Output()
+		out, err := exec.Command("/usr/bin/open", applicationPath).CombinedOutput()
 		if err != nil {
-			return fmt.Errorf("%s; %s", err, out)
+			return fmt.Errorf("%s; %s", err, string(out))
 		}
 		return nil
 	}
@@ -94,9 +94,9 @@ func (u *Updater) applyUpdate(localPath string) (err error) {
 
 	// Update spotlight, ignore (but log) errors
 	u.log.Debug("Updating spotlight: %s", destinationPath)
-	mdimportOut, mdierr := exec.Command("/usr/bin/mdimport", destinationPath).Output()
+	mdimportOut, mdierr := exec.Command("/usr/bin/mdimport", destinationPath).CombinedOutput()
 	if mdierr != nil {
-		u.log.Errorf("Error trying to update spotlight: %s; %s", mdierr, mdimportOut)
+		u.log.Errorf("Error trying to update spotlight: %s; %s", mdierr, string(mdimportOut))
 	}
 	return
 }
--- a/updater_osx.go
+++ b/updater_osx.go
@@ -73,12 +73,12 @@ func (u *Updater) openApplication(applicationPath string) error {
 	tryOpen := func() error {
 		out, err := exec.Command("/usr/bin/open", applicationPath).CombinedOutput()
 		if err != nil {
-			return fmt.Errorf("%s; %s", err, string(out))
+			return fmt.Errorf("Open error: %s; %s", err, string(out))
 		}
 		return nil
 	}
 	if err := tryOpen(); err != nil {
-		u.log.Errorf("Open error, trying again in a few seconds; %s", err)
+		u.log.Errorf("Open error (trying again in a few seconds): %s", err)
 		time.Sleep(3 * time.Second)
 		return tryOpen()
 	}
--- a/updater.go
+++ b/updater.go
@@ -710,20 +710,23 @@ func (u *Updater) restart(ctx Context, updateQuitResponse keybase1.UpdateQuitRes
 	}
 
 	u.log.Debug("App reported its PID as %d", updateQuitResponse.Pid)
-	p, err := os.FindProcess(updateQuitResponse.Pid)
-	if err != nil {
+	p, perr := os.FindProcess(updateQuitResponse.Pid)
+	if perr != nil {
+		err = fmt.Errorf("Error finding process: %s", perr)
 		return
 	}
 	u.log.Debug("Killing app")
-	err = p.Kill()
-	if err != nil {
+	kerr := p.Kill()
+	if kerr != nil {
+		err = fmt.Errorf("Error killing app: %s", kerr)
 		return
 	}
 	didQuit = true
 
 	u.log.Debug("Opening app at %s", updateQuitResponse.ApplicationPath)
-	err = openApplication(updateQuitResponse.ApplicationPath)
-	if err != nil {
+	oerr := u.openApplication(updateQuitResponse.ApplicationPath)
+	if oerr != nil {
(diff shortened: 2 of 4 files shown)


gabriel:
@cjb @chrisnojima

cjb:
We could merge this.  But we still don't have any theories for:

* why `open` failed
* why it just started failing yesterday after working forever

Any ideas on those?

updater_osx.go
```diff
@@ -68,9 +69,20 @@ func (u *Updater) checkPlatformSpecificUpdate(sourcePath string, destinationPath
...
 }
 
-func openApplication(applicationPath string) error {
-	_, err := exec.Command("/usr/bin/open", applicationPath).Output()
-	return err
+func (u *Updater) openApplication(applicationPath string) error {
+	tryOpen := func() error {
+		out, err := exec.Command("/usr/bin/open", applicationPath).CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("%s; %s", err, string(out))
```
maxtaco: Can the error message also say that `open` caused this problem too?

gabriel:
I digged a little bit trying to find the source code to /usr/bin/open to see if I could see under what scenarios a exit status of 1 would result... I pulled all the source for https://opensource.apple.com/ and grepped for strings present in the man page, but was unable to find it.

Just trying stuff out, I was able to get /usr/bin/open to return exit code 1 if:
- the file didn't exist
- if you tried to open from trash

I'm assuming there are other scenarios where open might exit with this code... maybe it always returns this on error... or maybe uses this code there is a file related issue...

If the application is already open then it will noop return exit code 0, so it seems like that isn't the issue. We do use SIGKILL on the app process.

This feels like a race condition or possibly flakiness arising from the OS (maybe Gatekeeper?) after a SIGKILL. I have seen Gatekeeper act weird. It seems maybe we shouldn't be using SIGKILL and it is too abrupt especially for something like Electron where it has other helper processes, etc?

I don't know why this would appear now, other than there have been lots of Electron patch releases lately so maybe some timing changed.

I looked at Sparkle and it uses [NSWorkspace openFile.](https://github.com/sparkle-project/Sparkle/blob/445cc55794a889aa7fa11c98887779b45992c56b/Sparkle/Autoupdate/Autoupdate.m#L219) Squirrel uses [NSWorkspace launchApplicationAtURL](https://github.com/Squirrel/Squirrel.Mac/blob/4a5041124602b061155e994429ed7958cae2411d/Squirrel/ShipIt-main.m#L126).

We could make our own open tool (or possibly cgo) which uses these library calls, but I'm not sure if it would give us anything.

gabriel:
We are also assuming (golang) os.Rename is atomic but maybe we shouldn't be... This would cause the /usr/bin/open to fail with exit code 1 (with file not found)...  

"There are systems where it cannot be done atomically, therefore it is not atomic."
https://groups.google.com/forum/#!topic/golang-nuts/ZjRWB8bMhv4

"os: make Rename atomic on Windows", 
"In Posix-compliant OSes, the os.Rename() function works properly (and atomically), including when replacing an existing file"
https://github.com/golang/go/issues/8914

maxtaco:
Is a directory or a file being swapped?  If a file, we can use `exchangedata(2)`

gabriel:
It's an (app bundle) directory.

cjb:
> We are also assuming (golang) os.Rename is atomic but maybe we shouldn't be... 

I think it's saying that if it's backed by `rename(2)`, and it's running on POSIX, then it *is* atomic, as POSIX requires.  But not every platform supported by golang is POSIX, so they can't make a general statement.

cjb:
I think the next step should be to run dtruss while reproing. It'll test our current hypothesis as well as letting us the see the error returned from `open`.

cjb:
I got dtruss going.  It isn't capturing everything going on, but this looks like what we were looking for:
```
  579/0x1900:  write_nocancel(0x2, "LSOpenURLsWithRole() failed with error -10810 for the file /Applications/Keybase.app.\0", 0x55)              = 85 0
```

cjb:
Good news: I'm now able to repro with a service that I built myself that includes this PR.

Bad news: The three second delay doesn't help.

```
2016-04-11T11:40:34.562989 ▶ [DEBU keybase updater.go:712] 1ea App reported its PID as 501
2016-04-11T11:40:34.563005 ▶ [DEBU keybase updater.go:717] 1eb Killing app
2016-04-11T11:40:34.563021 ▶ [DEBU keybase updater.go:724] 1ec Opening app at /Applications/Keybase.app
2016-04-11T11:40:34.570120 ▶ [DEBU keybase log.go:131] 1ed {unix} EOF
2016-04-11T11:40:34.570321 ▶ [DEBU keybase main.go:115] 1ee Handle() complete for connection 1
2016-04-11T11:40:34.579557 ▶ [DEBU keybase log.go:131] 1ef {unix} EOF
2016-04-11T11:40:34.579813 ▶ [DEBU keybase main.go:115] 1f0 Handle() complete for connection 3
2016-04-11T11:40:34.579932 ▶ [DEBU keybase log.go:131] 1f1 {unix} EOF
2016-04-11T11:40:34.580061 ▶ [DEBU keybase main.go:115] 1f2 Handle() complete for connection 2
2016-04-11T11:40:34.606877 ▶ [ERRO keybase standard.go:197] 1f3 Open error (trying again in a few seconds): Open error: exit status 1; LSOpenURLsWithRole() failed with error -10810 for the file /Applications/Keybase.app.
2016-04-11T11:40:37.649005 ▶ [DEBU keybase updater.go:735] 1f4 Cleaning up after update
2016-04-11T11:40:37.649066 ▶ [DEBU keybase updater.go:737] 1f5 Removing /var/folders/hg/f5t3mfpj3djbjr61h5x2pbs00000gp/T/KeybaseUpdates/Keybase-1.0.15-20160406130015+05efae8.zip.unzipped
2016-04-11T11:40:37.649173 ▶ [DEBU keybase updater.go:737] 1f6 Removing /var/folders/hg/f5t3mfpj3djbjr61h5x2pbs00000gp/T/KeybaseUpdates/Keybase-1.0.15-20160406130015+05efae8.zip
2016-04-11T11:40:37.651802 ▶ [DEBU keybase update.go:52] 1f7 - UpdateEngine Run
2016-04-11T11:40:37.651835 ▶ [DEBU keybase engine.go:66] 1f8 - RunEngine(Updater) -> ERROR: Error opening app: Open error: exit status 1; LSOpenURLsWithRole() failed with error -10810 for the file /Applications/Keybase.app.
2016-04-11T11:40:37.653652 ▶ [DEBU keybase log.go:131] 1f9 {unix} EOF
2016-04-11T11:40:37.653772 ▶ [DEBU keybase main.go:115] 1fa Handle() complete for connection 7
```

cjb:
I tried sending p.Signal(syscall.SIGTERM) instead of p.Kill (SIGKILL).  It changed the output but didn't help:
```
2016-04-11T11:58:00.970415 ▶ [ERRO keybase standard.go:197] 0c5 Open error (trying again in a few seconds): Open error: exit status 1; LSOpenURLsWithRole() failed with error -600 for the file /Applications/Keybase.app.
2016-04-11T11:58:04.010208 ▶ [DEBU keybase updater.go:736] 0c6 Cleaning up after update
2016-04-11T11:58:04.010275 ▶ [DEBU keybase updater.go:738] 0c7 Removing /var/folders/hg/f5t3mfpj3djbjr61h5x2pbs00000gp/T/KeybaseUpdates/Keybase-1.0.15-20160406130015+05efae8.zip.unzipped
2016-04-11T11:58:04.010399 ▶ [DEBU keybase updater.go:738] 0c8 Removing /var/folders/hg/f5t3mfpj3djbjr61h5x2pbs00000gp/T/KeybaseUpdates/Keybase-1.0.15-20160406130015+05efae8.zip
2016-04-11T11:58:04.013207 ▶ [DEBU keybase update.go:52] 0c9 - UpdateEngine Run
2016-04-11T11:58:04.013250 ▶ [DEBU keybase engine.go:66] 0ca - RunEngine(Updater) -> ERROR: Error opening app: Open error: exit status 1; LSOpenURLsWithRole() failed with error -10810 for the file /Applications/Keybase.app.
```
I'd guess that error -600 is because something's still held by the exiting old process.

cjb:
I tried waiting 9 seconds instead of 3 seconds, same result.

cjb:
We tried open()ing directly from the downloaded tempdir rather than moving the download into Applications and that worked.

cjb:
I tried killing the kbfs process during the 9 second delay between open() attempts, in case its open process from the "old" app bundle was preventing our relaunch, but it didn't help.

cjb:
Trying open with 10 one second delayed retries worked after 6 retries/opens.  But I think launchd is the parent process, not the service?  That's strange.

@chrisnojima suggests trying with gatekeeper turned off.

gabriel:
It might be a bug in Electron... we could try downgrading that to an earlier version?

cjb:
This is the diff that works for me:
```
diff --git a/go/vendor/github.com/keybase/go-updater/updater_osx.go b/go/vendor/github.com/keybase/go-updater/updater_osx.go
index 22312ae..271f806 100644
--- a/go/vendor/github.com/keybase/go-updater/updater_osx.go
+++ b/go/vendor/github.com/keybase/go-updater/updater_osx.go
@@ -6,6 +6,7 @@
 package updater
 
 import (
+       "fmt"
        "os"
        "os/exec"
        "os/user"
@@ -68,9 +69,23 @@ func (u *Updater) checkPlatformSpecificUpdate(sourcePath string, destinationPath
        return nil
 }
 
-func openApplication(applicationPath string) error {
-       _, err := exec.Command("/usr/bin/open", applicationPath).Output()
-       return err
+func (u *Updater) openApplication(applicationPath string) error {
+       tryOpen := func() error {
+               u.log.Errorf("Trying open")
+               out, err := exec.Command("/usr/bin/open", applicationPath).CombinedOutput()
+               if err != nil {
+                       u.log.Errorf("Open error: %s; %s", err, string(out))
+                       return fmt.Errorf("Open error: %s; %s", err, string(out))
+               }
+               return nil
+       }
+       for i := 0; i < 10; i++ {
+               if err := tryOpen(); err != nil {
+                       u.log.Errorf("Open error (trying again in a few seconds): %s", err)
+                       time.Sleep(1 * time.Second)
+               }
+       }
+       return nil
 }
 
 func (u *Updater) applyUpdate(localPath string) (err error) {
```

cjb:
I'm running my service outside of launchd (as far as launchd is concerned keybase.service is not running), so that might affect launchd's decisions.

gabriel:
oh i think i know the bug... it looks like the executable name changed from

/Applications/Keybase.app/Contents/MacOS/Electron to /Applications/Keybase.app/Contents/MacOS/Keybase

This causes Gatekeeper to get confused.

gabriel:
I ran into this a few months ago... I'm checking now to see if a electron-packager change did this

cjb:
Agree, the filename changed between good build and bad build.  Electron changed from v0.37.2 to v0.37.3.

PR #7, opened 2016-04-13
Run command with timeout

gabriel:
Also includes a run command for JSON output.

### DIFF
--- a/command.go
+++ b/command.go
@@ -30,6 +30,9 @@ func runCommand(command string, args []string, combinedOutput bool, timeout time
 	if command == "" {
 		return nil, nil, fmt.Errorf("No command")
 	}
+	if timeout < 0 {
+		return nil, nil, fmt.Errorf("Invalid timeout: %s", timeout)
+	}
 	cmd := exec.Command(command, args...)
 	if cmd == nil {
 		return nil, nil, fmt.Errorf("No command")
--- a/command.go
+++ b/command.go
@@ -4,6 +4,7 @@
 package updater
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -26,50 +27,54 @@ func timerWithTimeout(timeout time.Duration, cmd *exec.Cmd, log logging.Logger) 
 
 // RunCommand runs a command and returns the combined stdout/err output
 func RunCommand(command string, args []string, timeout time.Duration, log logging.Logger) (string, error) {
-	out, _, err := runCommand(command, args, timeout, log)
-	return out, err
+	out, _, err := runCommand(command, args, true, timeout, log)
+	if out == nil {
+		return "", err
+	}
+	return string(out), err
 }
 
-func runCommand(command string, args []string, timeout time.Duration, log logging.Logger) (string, *os.Process, error) {
+// runCommand runs a command and returns the output. If combinedOutput is true,
+// combined stdout/stderr output is returned, otherwise only stdout is returned.
+func runCommand(command string, args []string, combinedOutput bool, timeout time.Duration, log logging.Logger) ([]byte, *os.Process, error) {
 	log.Debugf("Command: %s %s", command, args)
+	if command == "" {
+		return nil, nil, fmt.Errorf("No command")
+	}
 	cmd := exec.Command(command, args...)
 	if cmd == nil {
-		return "", nil, fmt.Errorf("No command")
+		return nil, nil, fmt.Errorf("No command")
 	}
 	timer := timerWithTimeout(timeout, cmd, log)
 	if timer != nil {
 		defer timer.Stop()
 	}
-	out, err := cmd.CombinedOutput()
-	if err != nil {
-		return string(out), cmd.Process, fmt.Errorf("Error running command: %s", err)
+	var out []byte
+	var err error
+	if combinedOutput {
+		// Both stdout and stderr
+		out, err = cmd.CombinedOutput()
+	} else {
+		// Only stdout
+		out, err = cmd.Output()
 	}
-	return string(out), cmd.Process, nil
+	if err != nil {
+		return out, cmd.Process, fmt.Errorf("Error running command: %s", err)
+	}
+	return out, cmd.Process, nil
 }
 
 // RunJSONCommand runs a command (with timeout) expecting JSON output with result interface
 func RunJSONCommand(command string, args []string, result interface{}, timeout time.Duration, log logging.Logger) error {
-	log.Debugf("Command: %s %s", command, args)
-	cmd := exec.Command(command, args...)
-	if cmd == nil {
-		return fmt.Errorf("No command")
-	}
-	stdout, err := cmd.StdoutPipe()
+	out, _, err := runCommand(command, args, false, timeout, log)
 	if err != nil {
-		return fmt.Errorf("Error getting stdout: %s", err)
+		return err
 	}
-	if err := cmd.Start(); err != nil {
-		return fmt.Errorf("Error starting: %s", err)
+	if out == nil {
+		return fmt.Errorf("No output")
 	}
-	timer := timerWithTimeout(timeout, cmd, log)
-	if timer != nil {
-		defer timer.Stop()
-	}
-	if err := json.NewDecoder(stdout).Decode(&result); err != nil {
+	if err := json.NewDecoder(bytes.NewReader(out)).Decode(&result); err != nil {
 		return fmt.Errorf("Error in result: %s", err)
-	}
-	if err := cmd.Wait(); err != nil {
-		return fmt.Errorf("Error waiting: %s", err)
 	}
 	return nil
 }
--- a/command.go
+++ b/command.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"syscall"
 	"time"
 
 	"github.com/keybase/go-logging"
@@ -25,6 +26,8 @@ func RunCommand(command string, args []string, timeout time.Duration, log loggin
 
 // runCommand runs a command and returns the output. If combinedOutput is true,
 // combined stdout/stderr output is returned, otherwise only stdout is returned.
+// We will send TERM signal and wait 1 second or timeout, whichever is less,
+// before calling KILL.
 func runCommand(command string, args []string, combinedOutput bool, timeout time.Duration, log logging.Logger) ([]byte, *os.Process, error) {
 	log.Debugf("Command: %s %s", command, args)
 	if command == "" {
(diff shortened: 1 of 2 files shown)


chrisnojima:
@keybase/react-hackers

command.go
```diff
@@ -0,0 +1,75 @@
...
+func runCommand(command string, args []string, timeout time.Duration, log logging.Logger) (string, *os.Process, error) {
+	log.Debugf("Command: %s %s", command, args)
+	cmd := exec.Command(command, args...)
+	if cmd == nil {
+		return "", nil, fmt.Errorf("No command")
+	}
+	timer := timerWithTimeout(timeout, cmd, log)
+	if timer != nil {
+		defer timer.Stop()
+	}
```
chrisnojima: what happens if timer returns nil

command.go
```diff
@@ -0,0 +1,75 @@
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return string(out), cmd.Process, fmt.Errorf("Error running command: %s", err)
+	}
+	return string(out), cmd.Process, nil
+}
+
+// RunJSONCommand runs a command (with timeout) expecting JSON output with result interface
+func RunJSONCommand(command string, args []string, result interface{}, timeout time.Duration, log logging.Logger) error {
+	log.Debugf("Command: %s %s", command, args)
```
chrisnojima: shouldn't we have a more generic runCommand which outputs the stdout/stderr sep. then we can call it with RunCommand or RunJSONCommand, where RunCommand merges the output into a string and RunJSONCommand does the parsing back and forth?
just seems like there's a bunch of the same code

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/keybase/go-logging"
+)
+
+func TestEmptyRunCommand(t *testing.T) {
+	log := logging.Logger{Module: "test"}
+	_, err := RunCommand("", nil, time.Second, log)
```
chrisnojima: we should verify the first param is '' and not ignore it

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Logf("Error: %s", err)
+	if err == nil {
+		t.Fatal("Should have errored")
+	}
+}
+
+func TestInvalidRunCommand(t *testing.T) {
+	log := logging.Logger{Module: "test"}
+	_, err := RunCommand("invalidexecutable", nil, time.Second, log)
```
chrisnojima: we should verify the first param is '' and not ignore it

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+// this source code is governed by the included BSD license.
+
+package updater
+
+import (
+	"os"
+	"reflect"
+	"testing"
+	"time"
+
```
chrisnojima: what happens if time.Duration <= 0

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	_, err := RunCommand("invalidexecutable", nil, time.Second, log)
+	t.Logf("Error: %s", err)
+	if err == nil {
+		t.Fatal("Should have errored")
+	}
+}
+
+func TestRunCommandEcho(t *testing.T) {
+	log := logging.Logger{Module: "test"}
+	out, err := RunCommand("echo", []string{"arg1", "arg2"}, time.Second, log)
```
chrisnojima: where do we test the combined output. writing something to stdout and stderr

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	expected := "arg1 arg2\n"
+	if out != expected {
+		t.Errorf("Unexpected output: %q != %q", out, expected)
+	}
+}
+
+func TestRunCommandTimeout(t *testing.T) {
+	log := logging.Logger{Module: "test"}
+	_, err := RunCommand("sleep", []string{"10"}, time.Second, log)
```
chrisnojima: we should verify the first param is '' and not ignore it

chrisnojima: where are we actually testing the timeout. how do we know 1 second passed

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	if err == nil {
```
chrisnojima: where are we verifying we actually killed the process

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+		t.Fatal("Expected timeout error")
```
chrisnojima: where are we testing if kill fails

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+type testObj struct {
+	StringVar string `json:"stringVar"`
+	NumberVar int    `json:"numberVar"`
+	BoolVar   bool   `json:"boolVar"`
+}
+
+const testJSON = `{
```
chrisnojima: maybe this should have at least 1 level of nested object, like sub: {"hello": "world"}

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Logf("Out: %#v", testValOut)
+	if !reflect.DeepEqual(testVal, testValOut) {
+		t.Errorf("Invalid object: %#v", testValOut)
+	}
+}
+
+func TestRunJSONCommandTimeout(t *testing.T) {
+	log := logging.Logger{Module: "test"}
+	var testValOut testObj
+	err := RunJSONCommand("sleep", []string{"10"}, &testValOut, time.Second, log)
```
chrisnojima: if the run json code above has its own timing code and etc then we need all the tests from the above test also. checking the time actually was a second, that the process was killed, etc

command_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+// TestTimeoutProcessState checks to make sure process is killed after timeout
+func TestTimeoutProcessState(t *testing.T) {
+	log := logging.Logger{Module: "test"}
+	_, process, err := runCommand("sleep", []string{"10"}, time.Second, log)
+	if err == nil {
+		t.Fatal("Expected timeout error")
+	}
+	findProcess, _ := os.FindProcess(process.Pid)
+	perr := findProcess.Kill()
```
chrisnojima: will this error out if this Kill() itself just errored? is that possible

maxtaco:
@jzila can you help with golang review?

command.go
gabriel: If timer is nil, it means time.AfterFunc returned nil... The docs  https://golang.org/pkg/time/#AfterFunc don't say if this could happen.

Because that std library method returns a ptr, the check here is preventing a crash if it returns nil.

command.go
```diff
@@ -0,0 +1,75 @@
...
+
+func timerWithTimeout(timeout time.Duration, cmd *exec.Cmd, log logging.Logger) *time.Timer {
+	return time.AfterFunc(timeout, func() {
+		if cmd != nil && cmd.Process != nil {
+			log.Warningf("Command timed out, killing")
+			if err := cmd.Process.Kill(); err != nil {
+				log.Warningf("Error trying to kill process: %s", err)
+			}
+		}
+	})
```
jzila: What is this supposed to be doing? Right now, it's never calling `cmd`. After `timeout`, it checks to see if cmd is running, and claiming it timed out if it's running (although nothing ran it). I think what you're actually looking for is:
```go
func runWithTimeout(timeout time.Duration, cmd *exec.Cmd, log logging.Logger) {
    cmd.Start()
    doneCh := make(chan struct{})
    go func() {
        cmd.Wait()
        close(doneCh)
    }
    select {
    case <-doneCh:
    case <-time.After(timeout):
        log.Warningf("Command timed out, killing")
        if err := cmd.Process.Kill(); err != nil {
            log.Warningf("Error trying to kill process: %s", err)
        }
    }
}
```

jzila: If you want it to be async then you can return an error channel instead of an error, and run the entire second half of this function in a goroutine which writes to a buffered error channel.

command.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	timer := timerWithTimeout(timeout, cmd, log)
+	if timer != nil {
+		defer timer.Stop()
+	}
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return string(out), cmd.Process, fmt.Errorf("Error running command: %s", err)
+	}
+	return string(out), cmd.Process, nil
```
jzila: Why are we returning `cmd.Process`?

command_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"testing"
+	"time"
+
+	"github.com/keybase/go-logging"
+)
+
+var log = logging.Logger{Module: "test"}
+
+func TestEmptyRunCommand(t *testing.T) {
+	out, err := RunCommand("", nil, time.Second, log)
```
jzila: We should be testing all timeouts with [clockwork](/jonboulle/clockwork) so it doesn't block the tests.

command.go
gabriel: I had a different impl cause I was streaming output into decoder, but we don't need that.. so I unified them, RunJSONCommand now calls runCommand.

command_test.go
gabriel: Fixed.

command_test.go
gabriel: Fixed.

command_test.go
gabriel: A negative duration means the timeout would never happen. I created a test for this.

command_test.go
gabriel: so exec.Cmd CombinedOutput() is std library function which says it gives combined output... So I don't think we need to test that.

command_test.go
gabriel: Added check that at least a second elapsed in TestRunCommandTimeout

command_test.go
gabriel: TestTimeoutProcessKilled tries to verify that the process was killed...

Unfortunately go doesn't expose the state of a process unless you call Wait(), but Wait() error if it's not running, so I am calling kill and asserting an error occurs since you can't kill a non-existent process.

command_test.go
```diff
@@ -0,0 +1,158 @@
...
+	if out != "" {
+		t.Errorf("Unexpected output: %s", out)
```
jzila: Please use [testify](/stretchr/testify) to make asserts. This also trivializes JSON testing because it allows you to directly [assert JSON equality](https://godoc.org/github.com/stretchr/testify/require#JSONEq)

command_test.go
gabriel: If kill fails we log an error, but there is nothing really we can do...

command_test.go
gabriel: Added nested object

command_test.go
gabriel: JSON command now shared code

command_test.go
gabriel: This asserts that an error occurred so any error will cause it to pass. I think this is ok given my comment above.

command_test.go
chrisnojima: because we rely on some library to work doens't mean our code actually uses this right? if the intention is that this grabs and concats stderr/stdout we should test that so if we like don't use CombinedOutput() we'd catch that

command.go
```diff
@@ -0,0 +1,80 @@
...
+		// Both stdout and stderr
+		out, err = cmd.CombinedOutput()
+	} else {
+		// Only stdout
+		out, err = cmd.Output()
+	}
+	if err != nil {
+		return out, cmd.Process, fmt.Errorf("Error running command: %s", err)
+	}
+	return out, cmd.Process, nil
```
jzila: Why are we returning `cmd.Process`?

gabriel: This is so tests can assert that process was killed (as much as it can)

command.go
gabriel: Calling Output or CombinedOutput, starts, waits and gets output all in a single call. I think I prefer using a timer than doing Start() and then Wait() with channel code?

As far as being graceful...

> Everything will probably be just fine if you kill-9 something. If a program fails permanently and dramatically when it's kill-9'd, you should "remove it from your filesystem", because it can't handle other, unavoidable abrupt failures either.

Also there are problems with your kill code... you are assuming the process object is always there and if the process dies after TERM timeout it will error with Process.Kill() even though it got killed.

command_test.go
gabriel: Because I'm actually calling sleep executable, it would be hard to do this.

The tests take 4 seconds to run.. I'm ok with it being more of a real test environment without a fake clock.

command.go
jzila: Yeah I typed it in a text box so it's not 100%. I'm sure you could get around the issue you mentioned :). I'd definitely prefer to be explicit using channel code to coordinate rather than handling some of the code inside a timer, then stopping the timer elsewhere. Right now the coordinationg code is sprinkled between `CombinedOutput`, `timerWithTimeout`, and `runCommand`, and it's very difficult to follow what's happening. With my suggestion you'd have all the coordination in one place and the code flows linearly.

command_test.go
jzila: Yes but if the tests currently run in 500ms, you can run them on every save. Adding 4 seconds makes that impossible. We should really try to avoid adding sleeps to tests if at all possible, especially for such workload-light tests.

command_test.go
gabriel: Using testify asserts. Still using DeepEqual though since it's testing decoded object

jzila: order matters with `DeepEqual` but not with `JSONEq`. I've had random test failures with `DeepEqual` based on order of key insertion.

command_test.go
gabriel: Ah good point, will fix

command.go
gabriel: Yeah I removed the separate method, so it's inline now so it should be more readable.

So actually capturing output without using Output() or CombinedOutput requires a bunch of new code..e.g. calling Start(), Wait() then reading from the pipes.

I don't know the timer code seems legit and simple to me?

gabriel: Also I think the timer code has more desired behavior... either I get all the output in the time specified or i run kill... if kill occurs after it ends, then it doesn't error or crash.

jzila: Right so my suggestion already does the `Start` and `Wait`, which make the concurrency explicit against the timeout code, instead of in addition to. I've edited the comment to also give you output.

jzila: Regarding your desired behavior, these are all minor changes to the function I've proposed above. You don't have to return an error if the Kill() fails.

Graceful termination is always preferable to kill -9. There are many scenarios where a badly terminated process can cause misbehavior of future calls. You can leave opened file handles in a bad state, you can leave orphaned children, etc.

command_test.go
gabriel: Ok, got the tests down to 30ms and not testing with a fake clock.. I think that's preferable.

command_test.go
gabriel: Changed it to error on negative duration since linux and darwin differ on how they handle it

gabriel:
ok @jzila I think it's all good now

command.go
```diff
@@ -0,0 +1,104 @@
...
+	// Wait a second or timeout for termination, whichever less
+	termWait := time.Second
+	if timeout < termWait {
+		termWait = timeout
+	}
+	log.Warningf("Command timed out, terminating (will wait %s before killing)", termWait)
+	cmd.Process.Signal(syscall.SIGTERM)
+	select {
+	case <-doneCh:
+		log.Warningf("Terminated")
```
jzila: `log.Debugf` here. This is the expected outcome in most cases.

jzila: Oops nevermind, I scrolled too far.

jzila:
LGTM!

PR #9, opened 2016-04-13
Protocol for updater (v2)

gabriel:
Instead of vendoring and using avdl, it's best to decouple. You'll notice there isn't much required in the new protocol.

There will be a later PR for changing to use the new protocol.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -1,0 +1,119 @@
+// Copyright 2015 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package updater
+
+import "time"
+
+// Asset describes a downloadable file
+type Asset struct {
+	Name      string `codec:"name" json:"name"`
+	URL       string `codec:"url" json:"url"`
+	Digest    string `codec:"digest" json:"digest"`
+	Signature string `codec:"signature" json:"signature"`
+	LocalPath string `codec:"localPath" json:"localPath"`
+}
+
+// UpdateType is the update type.
+// This is an int type for compatibility.
+type UpdateType int
+
+const (
+	// UpdateTypeNormal is a normal update
+	UpdateTypeNormal UpdateType = 0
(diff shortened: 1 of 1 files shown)


chrisnojima:
@keybase/react-hackers

chrisnojima:
logically seems fine, need golang review

maxtaco:
@jzila can you help with golang review?

protocol.go
```diff
@@ -0,0 +1,119 @@
...
+// UpdatePromptResponse is the result for UpdatePrompt
+type UpdatePromptResponse struct {
+	Action     UpdateAction `json:"action"`
+	AutoUpdate bool         `json:"autoUpdate"`
+}
+
+// UpdatePromptUI is a prompt interface
+type UpdatePromptUI interface {
+	// UpdatePrompt prompts for an update
+	UpdatePrompt(Update, UpdateOptions, UpdatePromptOptions) (*UpdatePromptResponse, error)
```
jzila: why does this return a pointer type?

protocol.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Time is milliseconds since epoch
+type Time int64
```
jzila: I don't think this type is helpful. `time.Unix()` takes an int64, and `time.Time.Unix()` returns an int64, so having a separate time here just adds casting complexity.

jzila:
LGTM after nits

protocol.go
gabriel: Wanted a nil response if the user didn't actually response (like on an error) instead of empty object.

jzila: The standard Go way of doing this is to return a zero value `UpdatePromptResponse{}`

protocol.go
gabriel: Oh right... yeah I pulled this in from the avdl protocol side without thinking that I could nuke it instead... removed.

PR #10, opened 2016-04-13
Saltpack verify (for v2)

gabriel:
This is from the client code but only has a dependency on keybase/saltpack.

New code will use this to saltpack verify in the future.


chrisnojima:
@keybase/react-hackers

chrisnojima:
@oconnor663 maybe you can take a peek at this

saltpack.go
```diff
@@ -0,0 +1,135 @@
...
+type KID string
+
+const (
+	// KIDLen is KID length in bytes
+	KIDLen = 35
+	// KIDSuffix is KID suffix (byte)
+	KIDSuffix = 0x0a
+	// KIDVersion is current version of KID
+	KIDVersion = 0x1
+)
```
oconnor663: It looks like we have to duplicate these constants from libkb, because we specify the set of valid libkb-style KIDs, rather than just using a set of raw signing public keys. Any reason not to just use `"9092ae4e790763dc7343851b977930f35b16cf43ab0ad900a2af3d3ad5cea1a1"` instead of `"01209092ae4e790763dc7343851b977930f35b16cf43ab0ad900a2af3d3ad5cea1a10a"`, and avoid the KID-munging code?

saltpack.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func kidIsIn(k KID, list []KID) bool {
+	for _, h := range list {
+		if h.Equal(k) {
+			return true
+		}
+	}
+	return false
+}
```
oconnor663: Maybe pass the kids in a `map[KID]struct{}` instead of a slice, and avoid needing to write this function?

saltpack.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if err = checkSender(skey); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type keyring struct{}
```
oconnor663: I think there's a simple keyring implemented somewhere in the `saltpack` lib itself.

oconnor663: Yeah would `github.com/keybase/saltpack/basic::NewKeyring()` work here?

saltpack.go
gabriel: Oh yeah that removes a bunch of that code.. thx!

saltpack.go
gabriel: ok yeah, using map[string]bool

saltpack.go
gabriel: Am I allowed to use basic... I was under the impression from code review in past that that was only a example implementation and shouldn't be used directly.

gabriel:
@oconnor663 i think all good now?

saltpack.go
maxtaco: Sorry, sounds like something I might have mistakenly said!

maxtaco: I apologize if that was the wrong idea.  I vaguely remember this conversation

oconnor663:
I think this is a good place to use the basic one, since Max doesn't object. Other than that, LGTM.

PR #11, opened 2016-04-14
Util (v2)

gabriel:
These are some util dependencies from client for v2 updater. 

The config will use the SafeWriter (file.go).
The http will use the DiscardAndCloseBody (http.go).

I've added some tests too.

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -4,7 +4,6 @@
 package util
 
 import (
-	"fmt"
 	"io"
 	"os"
 
@@ -31,7 +30,7 @@ func NewFile(name string, data []byte, perm os.FileMode) File {
 
 // Save file
 func (f File) Save(log logging.Logger) error {
-	return SafeWriteToFile(f, f.perm, log)
+	return safeWriteToFile(f, f.perm, log)
 }
 
 // GetFilename returns the file name for SafeWriter
@@ -45,8 +44,8 @@ func (f File) WriteTo(w io.Writer) (int64, error) {
 	return int64(n), err
 }
 
-// SafeWriteToFile to safely write to a file. Use mode=0 for default permissions.
-func SafeWriteToFile(t SafeWriter, mode os.FileMode, log logging.Logger) error {
+// safeWriteToFile to safely write to a file. Use mode=0 for default permissions.
+func safeWriteToFile(t SafeWriter, mode os.FileMode, log logging.Logger) error {
 	fn := t.GetFilename()
 	log.Debugf("Writing to %s", fn)
 	tmpfn, tmp, err := OpenTempFile(fn, "", mode)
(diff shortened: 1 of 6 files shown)


chrisnojima:
@keybase/react-hackers

util/file.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// OpenTempFile creates an opened termporary file. Use mode=0 for default
+// permission (0600).
+//
+//   OpenTempFile("foo", ".zip", 0755) => "foo.RCG2KUSCGYOO3PCKNWQHBOXBKACOPIKL.zip"
+//   OpenTempFile(path.Join(os.TempDir(), "foo"), "", 0) => "/tmp/foo.RCG2KUSCGYOO3PCKNWQHBOXBKACOPIKL"
+//
+func OpenTempFile(prefix string, suffix string, mode os.FileMode) (string, *os.File, error) {
+	if prefix == "" {
```
chrisnojima: curious why prefix is required but suffix isn't

util/file.go
```diff
@@ -0,0 +1,99 @@
...
+func SafeWriteToFile(t SafeWriter, mode os.FileMode, log logging.Logger) error {
+	fn := t.GetFilename()
+	log.Debugf("Writing to %s", fn)
+	tmpfn, tmp, err := OpenTempFile(fn, "", mode)
+	log.Debugf("Temporary file generated: %s", tmpfn)
+	if err != nil {
+		return err
+	}
+
+	_, err = t.WriteTo(tmp)
```
chrisnojima: does it make sense to defer the tmp.Close() up here and not below in the various logic flows

chrisnojima:
needs golang review

maxtaco:
@jzila can you take a go-y look?

util/file.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// WriteTo is for SafeWriter
+func (f File) WriteTo(w io.Writer) (int64, error) {
+	n, err := w.Write(f.data)
+	return int64(n), err
+}
+
+// SafeWriteToFile to safely write to a file. Use mode=0 for default permissions.
+func SafeWriteToFile(t SafeWriter, mode os.FileMode, log logging.Logger) error {
```
jzila: does this need to be exported? Looks like it's only called from `File.Save()`

util/file.go
```diff
@@ -0,0 +1,99 @@
...
+	return err
+}
+
+// OpenTempFile creates an opened termporary file. Use mode=0 for default
+// permission (0600).
+//
+//   OpenTempFile("foo", ".zip", 0755) => "foo.RCG2KUSCGYOO3PCKNWQHBOXBKACOPIKL.zip"
+//   OpenTempFile(path.Join(os.TempDir(), "foo"), "", 0) => "/tmp/foo.RCG2KUSCGYOO3PCKNWQHBOXBKACOPIKL"
+//
+func OpenTempFile(prefix string, suffix string, mode os.FileMode) (string, *os.File, error) {
```
jzila: Does this need to be exported?

util/file.go
```diff
@@ -0,0 +1,99 @@
...
+	log.Debugf("Temporary file generated: %s", tmpfn)
+	if err != nil {
+		return err
+	}
+
+	_, err = t.WriteTo(tmp)
+	if err == nil {
+		err = tmp.Close()
+		if err == nil {
+			err = os.Rename(tmpfn, fn)
```
jzila: what if the target filename exists?

util/file.go
```diff
@@ -0,0 +1,99 @@
...
+	fn := t.GetFilename()
+	log.Debugf("Writing to %s", fn)
+	tmpfn, tmp, err := OpenTempFile(fn, "", mode)
+	log.Debugf("Temporary file generated: %s", tmpfn)
+	if err != nil {
+		return err
+	}
+
+	_, err = t.WriteTo(tmp)
+	if err == nil {
```
jzila: this would be better laid out as follows:
```go
if err != nil {
    log.Errorf("Error writing temporary file %s: %s", tmpfn, err)
    _ = tmp.Close()
    _ = os.Remove(tmpfn)
    return err
}
err = tmp.Close()
if err != nil {
    log.Errorf("Error closing temporary file %s: %s", tmpfn, err)
    _ = os.Remove(tmpfn)
    return err
}
log.Debugf("Wrote to %s", fn)
return os.Rename(tmpfn, fn) // TODO: rollback if this fails
```

util/util.go
```diff
@@ -0,0 +1,43 @@
...
+	return strings.Join(arrNew, delimeter)
+}
+
+// RandString returns random (base32) string with prefix.
+func RandString(prefix string, numbytes int) (string, error) {
+	buf, err := RandBytes(numbytes)
+	if err != nil {
+		return "", err
+	}
+	str := base32.StdEncoding.EncodeToString(buf)
```
jzila: why base32 and not base64?

maxtaco: this seems safer due to non-a-zA-Z0-9 characters in base64

util/util.go
```diff
@@ -0,0 +1,43 @@
...
+	if prefix != "" {
+		str = strings.Join([]string{prefix, str}, "")
```
jzila: I think this would be cleaner as:
```go
str := fmt.Sprintf("%s%s", prefix, base32.StdEncoding.EncodeToString(buf))
return str, nil
```

jzila:
LGTM after nits

util/file.go
gabriel: yeah this will be used in the future by another PR

util/util.go
gabriel: yep thx... made it `str := prefix + base32.StdEncoding.EncodeToString(buf)`

util/file.go
gabriel: Yep... i copied from client, but good to fix it up...

Added rename error check so now it's:

```
	_, err = t.WriteTo(tmp)
	if err != nil {
		log.Errorf("Error writing temporary file %s: %s", tmpfn, err)
		_ = tmp.Close()
		_ = os.Remove(tmpfn)
		return err
	}
	err = tmp.Close()
	if err != nil {
		log.Errorf("Error closing temporary file %s: %s", tmpfn, err)
		_ = os.Remove(tmpfn)
		return err
	}
	err = os.Rename(tmpfn, fn)
	if err != nil {
		log.Errorf("Error renaming temporary file %s to %s: %s", tmpfn, fn, err)
		_ = os.Remove(tmpfn)
		return err
	}
	log.Debugf("Wrote to %s", fn)
	return nil
```

util/file.go
gabriel: Fixed

util/file.go
gabriel: not in this repo... made it private

util/file.go
gabriel: we need to close the file before renaming it

util/file.go
gabriel: sure...  removed prefix requirement

PR #13, opened 2016-04-14
Update prompt app for OS X

gabriel:
Separated into 2 commits for easier reviewing. You can ignore the project files. There are no dependencies.

It looks something like this (mock version):
2016-04-13 at 6 06 pm


chrisnojima:
@keybase/react-hackers

osx/Updater/AppDelegate.m
```diff
@@ -0,0 +1,56 @@
+//
+//  AppDelegate.m
+//  Updater
+//
+//  Created by Gabriel on 4/7/16.
```
chrisnojima: should likely change your xcode settings to use keybase

osx/Updater/AppDelegate.m
```diff
@@ -0,0 +1,56 @@
...
+  if (args.count < 1) {
+    [self exitWithError:KBMakeError(@"No args for process")];
+  }
+
+  NSArray *subargs = [args subarrayWithRange:NSMakeRange(1, args.count-1)];
+  if (subargs.count < 1) {
+    [self exitWithError:KBMakeError(@"No args")];
+  }
+
+  if (![subargs[0] isKindOfClass:NSString.class]) {
```
chrisnojima: isn't this guaranteed to be a string? https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSProcessInfo_Class/#//apple_ref/occ/instp/NSProcessInfo/arguments

osx/Updater/AppDelegate.m
```diff
@@ -0,0 +1,56 @@
...
+
+- (void)applicationDidFinishLaunching:(NSNotification *)notification {
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [self run];
+  });
+}
+
+- (void)exitWithError:(NSError *)error {
+  NSLog(@"Error: %@", error.localizedDescription);
+  fflush(stderr);
```
chrisnojima: since we're taking in stdin and stderr, should we just flush both always

osx/Updater/Prompt.m
```diff
@@ -0,0 +1,118 @@
...
+  if (![input isKindOfClass:[NSDictionary class]]) {
+    completion(KBMakeError(@"Invalid input"), nil);
+    return;
+  }
+
+  [self showUpdatePrompt:input presenter:presenter completion:completion];
+}
+
++ (void)showUpdatePrompt:(NSDictionary *)input presenter:(NSModalResponse (^)(NSAlert *alert))presenter completion:(void (^)(NSError *error, NSData *output))completion {
+  NSString *title = [input objectForKey:@"title"];
```
chrisnojima: All these lines assume a string is returned. If you make a json like
```
./Updater "{\"title\":\"world\", \"message\":\"a messageeeeee\", \"description\":\"a desc\", \"autoUpdate\":[true]}"
```

it'll die. same with null => [NSNull] and dictionaries and etc.  i think you'll need a class check to verify its really an NSString

chrisnojima: also we should have tests to check this

osx/Updater/Base.lproj/MainMenu.xib
```diff
@@ -0,0 +1,667 @@
...
+        <customObject id="-2" userLabel="File's Owner" customClass="NSApplication">
+            <connections>
+                <outlet property="delegate" destination="Voe-Tx-rLC" id="GzC-gU-4Uq"/>
+            </connections>
+        </customObject>
+        <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
+        <customObject id="-3" userLabel="Application" customClass="NSObject"/>
+        <customObject id="Voe-Tx-rLC" customClass="AppDelegate"/>
+        <customObject id="YLy-65-1bz" customClass="NSFontManager"/>
+        <menu title="Main Menu" systemMenu="main" id="AYu-sK-qS6">
```
chrisnojima: We should trim this menu

osx/Updater/Prompt.m
```diff
@@ -0,0 +1,118 @@
...
+    action = @"update";
+  } else if (response == NSAlertSecondButtonReturn) {
+    action = @"ignore";
+  }
+  NSLog(@"Action: %@", action);
+
+  NSError *error = nil;
+  NSDictionary *result = @{
+                           @"action": action,
+                           @"autoUpdate": autoCheckbox.state == NSOnState ? @YES : @NO,
```
chrisnojima: we should likely ignore auto update if they click ignore

osx/UpdaterTests/PromptTests.m
```diff
@@ -0,0 +1,40 @@
...
+#import <XCTest/XCTest.h>
+
+#import "Prompt.h"
+
+@interface PromptTests : XCTestCase
+@end
+
+@implementation PromptTests
+
+- (void)testUpdatePrompt {
```
chrisnojima: we should test bad json

chrisnojima: we should test autoUpdate setting being plumbed through correctly

chrisnojima:
dunno if we care but there's no sanity checking of inputs to the non-scrollable fields, so you can do something that'll kill your buttons

screen shot 2016-04-14 at 3 03 56 pm

chrisnojima:
actually we might care since we disable command-q quitting of this modal

osx/Updater/AppDelegate.m
gabriel: Changed org in project and updated files.

osx/Updater/AppDelegate.m
gabriel: Yep removed.

osx/Updater/AppDelegate.m
gabriel: Yep refactored to have a single exit method. Out and err are flushed on all exits.

osx/Updater/Prompt.m
gabriel: Did some refactoring. Added tests for invalid JSON types as values and as root.

I also refactored to have it continue, even with failed inputs instead of exiting with a non-zero status.

osx/Updater/Base.lproj/MainMenu.xib
gabriel: The default seems fine to me. There are a lot of keyboard shortcuts embedded in places in the menu that removing might cause them to break. Probably best to leave defaults, there isn't really a downside?

osx/Updater/Prompt.m
gabriel: Ah yep.. Fixed this... Also have a branch with test for this on the go side as well.

osx/UpdaterTests/PromptTests.m
gabriel: Added test for bad json, and adding assertions for output.

gabriel:
RE: Really big height.. since we're using an NSAlert which is not a window, there is no maxSize property. We could do an NSPanel but would have to do more UI code.

Maybe we can punt on this till the end and decide then?

gabriel:
OK, I think I got all the comments... this might be ready for final review

osx/Updater/Base.lproj/MainMenu.xib
chrisnojima: I think the menu options go counter to us having a smaller surface area. This just allows the users to do weird shit, like i can turn on the input box ruler and mess with stuff if i want to.

screen shot 2016-04-20 at 9 38 31 am

chrisnojima:
re: height we could just limit how long the string can possibly be. like 1000 chars or something

chrisnojima:
👍 . it looks like when it pops up when i run it from the command line its not in the foreground, which used to work?

gabriel:
Limiting title and message (non scrollable fields) to 700 chars.

Fixed appearing in background.

PR #14, opened 2016-04-15
Simplifying update sources, new keybase.io (api) source (v2)

gabriel:
This is the update source that makes request to keybase.io

Removes the sources prerelease tag which was confusing (and currently always on).


gabriel:
@keybase/updater-hackers

keybase/source.go
```diff
@@ -0,0 +1,104 @@
...
+
+// UpdateSource finds releases/updates on keybase.io
+type UpdateSource struct {
+	log      logging.Logger
+	endpoint string
+}
+
+// NewUpdateSource contructs an update source for keybase.io
+func NewUpdateSource(log logging.Logger) UpdateSource {
+	return newUpdateSource("https://keybase.io/_/api/1.0/pkg/update.json", log)
```
chrisnojima: should this be api.keybase.io?

keybase/source.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+	urlValues := url.Values{}
+	urlValues.Add("install_id", options.InstallID)
+	urlValues.Add("version", options.Version)
+	urlValues.Add("platform", options.Platform)
+	urlValues.Add("run_mode", options.Env)
+	urlValues.Add("os_version", options.OSVersion)
+	urlValues.Add("upd_version", options.UpdaterVersion)
+	u.RawQuery = urlValues.Encode()
```
chrisnojima: does this err out?

chrisnojima:
@jzila

keybase/source.go
```diff
@@ -0,0 +1,104 @@
...
+	urlString := u.String()
+
+	req, err := http.NewRequest("GET", urlString, nil)
+	client := &http.Client{
+		Timeout: time.Minute,
+	}
+	k.log.Infof("Request %#v", urlString)
+	resp, err := client.Do(req)
+	defer func() { _ = util.DiscardAndCloseBody(resp) }()
```
jzila: why does this need to be in a separate closure instead of `defer util.DiscardAndCloseBody(resp)`?

remote.go
```diff
@@ -0,0 +1,76 @@
...
+// FindUpdate returns update for options
+func (r RemoteUpdateSource) FindUpdate(options UpdateOptions) (update *Update, err error) {
+	sourceURL := r.sourceURL(options)
+	req, err := http.NewRequest("GET", sourceURL, nil)
+	client := &http.Client{
+		Timeout: time.Minute,
+	}
+	r.log.Infof("Request %#v", sourceURL)
+	resp, err := client.Do(req)
+	defer func() { _ = util.DiscardAndCloseBody(resp) }()
```
jzila: here too

keybase/source.go
```diff
@@ -0,0 +1,104 @@
...
+
+	var reader io.Reader = resp.Body
+	var res updateResponse
+	if err = json.NewDecoder(reader).Decode(&res); err != nil {
+		err = fmt.Errorf("Invalid API response %s", err)
+		return
+	}
+
+	if res.Status.Code != 0 {
+		err = fmt.Errorf("API returned error response: %#v", res)
```
jzila: you can return here and avoid the else

local.go
```diff
@@ -35,7 +38,7 @@ func digest(URL string) (digest string, err error) {
 	if err != nil {
 		return
 	}
-	defer f.Close()
+	defer func() { _ = f.Close() }()
```
jzila: why put this into a new function?

local.go
```diff
@@ -50,16 +53,19 @@ func readFile(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer sigFile.Close()
+	defer func() { _ = sigFile.Close() }()
```
jzila: why put this into a new function?

remote.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+// Description returns update source description
+func (r RemoteUpdateSource) Description() string {
+	return "Remote"
+}
+
+func (r RemoteUpdateSource) sourceURL(options UpdateOptions) string {
+	params := util.JoinPredicate([]string{options.Platform, options.Env, options.Channel}, "-", func(s string) bool { return s != "" })
```
jzila: can't this result in an ambiguous URI? `[]string{"foo", "", "foo"}` will resolve to the same URI as `[]string{"", "foo", "foo"}`.

remote.go
```diff
@@ -0,0 +1,76 @@
...
+		return
+	}
+
+	var reader io.Reader = resp.Body
+	var obj Update
+	if err = json.NewDecoder(reader).Decode(&obj); err != nil {
+		err = fmt.Errorf("Bad updater remote response %s", err)
+		return
+	}
+	update = &obj
```
jzila: probably cleaner just to do:
```go
r.log.Debugf(...)
return &obj, nil
```

jzila:
Finished first pass

local.go
gabriel: gometalinter! This code passes all go linters possible, one of which checks that err is not handled

remote.go
gabriel: yeah that's ok though

keybase/source.go
gabriel: That requires adding our cert... created story for that here: https://keybase.atlassian.net/browse/DESKTOP-925

local.go
jzila: -_- seems a little ridiculous to create a closure to satisfy a linter...

keybase/source.go
gabriel: See gometalinter comment

remote.go
gabriel: See gometalinter comment

local.go
gabriel: See gometalinter comment

remote.go
gabriel: Yeah cleaned this up a bit

keybase/source.go
gabriel: Yeah cleaned this up a bit

jzila:
I don't love the closure, but if it makes you happy to pass the metalinter then I'll allow it. :shipit:

gabriel:
Yeah those are weird... I added util methods with no error return value defined and then fixed those defer calls

local.go
gabriel: The errcheck linter is a really good one to satisfy though :)

But yeah, adding util methods that don't return error and using those is a better way to satisfy the errcheck and say, I don't care if it errors.

PR #15, opened 2016-04-15
Config for v2

gabriel:
Includes config, context and some platform specific code for config.

Context is keybase specific behavior for the updater.


gabriel:
@keybase/updater-hackers

keybase/config.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return err
+	}
+	if file == nil {
+		return fmt.Errorf("No file")
+	}
+	defer func() { _ = file.Close() }()
+
+	decoder := json.NewDecoder(file)
+	if err := decoder.Decode(&c.store); err != nil {
```
chrisnojima: will this write partial stuff to the store? like as its decoding the json will it be filling values in here until it finds an err? Is that what we want

keybase/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+	path, err := cfg.path()
+	assert.Nil(t, err, "%s", err)
+	assert.NotEqual(t, path, "", "No config path")
+
+	configDir, err := cfg.dir()
+	assert.Nil(t, err, "%s", err)
+	assert.NotEqual(t, configDir, "", "Config dir empty")
+	defer func() { _ = os.Remove(configDir) }()
+
+	err = cfg.SetUpdateAuto(false)
```
chrisnojima: we need a test to read the default store values before we set some

keybase/context.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// BeforeApply is called before an update is applied
+func (c context) BeforeApply() error {
+	return nil
+}
+
+// AfterApply is called before an update is applied
+func (c context) AfterApply() error {
+	output, err := updater.RunCommand(c.config.pathToKeybase, []string{"update", "notify", "after-apply"}, 10*time.Second, c.log)
```
chrisnojima: so in the psuedocode this is an async call and not the timeout call. i think its fine to use the timeout one, we should discuss but is 10 seconds enough? @maxtaco

keybase/platform_darwin.go
```diff
@@ -0,0 +1,50 @@
...
+	"strings"
+	"time"
+
+	"github.com/kardianos/osext"
+	"github.com/keybase/go-updater"
+)
+
+// destinationPath returns the app bundle path where this executable is located.
+// Currently we only support `/Applications/Keybase.app`.
+func (c config) destinationPath() string {
```
chrisnojima: What happens if its not installed there. Why can't we just find this relative to 'ourselves' as we'll always be inside the bundle anyways?

util/file.go
```diff
@@ -112,3 +113,33 @@ func OpenTempFile(prefix string, suffix string, mode os.FileMode) (string, *os.F
...
+func MakeParentDirs(path string) error {
+	// 2nd return value here is filename (not an error), which is not needed
+	dir, _ := filepath.Split(path)
+	exists, err := FileExists(dir)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		err = os.MkdirAll(dir, 0700)
```
chrisnojima: why 0700 vs 0600 for the file?

util/file_test.go
```diff
@@ -23,3 +24,18 @@ func TestNewFile(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestMakeParentDirs(t *testing.T) {
+	dir := filepath.Join(os.TempDir(), "TestMakeParentDirs")
```
chrisnojima: should test making a dir that already exists

chrisnojima:
@jzila

keybase/config.go
gabriel: Yeah it is streaming so it will mutate the store as it parses...

This is a good point, perhaps we want to to revert to the default state in case of any parse error.

I made this change and added a test for this.

We can change our mind though if we want config to mutate while parsing and we decide we want all valid state we can get?

keybase/context.go
gabriel: Yeah, we could bump the timeout to something large... maybe even something like 2 minutes?

keybase/platform_darwin.go
gabriel: Yeah we can support any app bundle. Fixed and added a test to get app bundle from path instead of whitelisting.

util/file_test.go
gabriel: Added test

util/file.go
gabriel: Its a directory so needs to be executable.

Added tests for checking file mode.

gabriel:
This is ready for review :)

keybase/config.go
```diff
@@ -0,0 +1,162 @@
...
+	log logging.Logger
+	// store is the config values
+	store store
+}
+
+// store is the config values
+type store struct {
+	InstallID string `json:"installId"`
+	Auto      bool   `json:"auto"`
+	AutoSet   bool   `json:"autoSet"`
```
maxtaco: What's the difference between `AutoSet` and `Auto`?

keybase/config.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	if file == nil {
+		return fmt.Errorf("No file")
+	}
+	defer util.Close(file)
+
+	decoder := json.NewDecoder(file)
+	var decodeStore store
+	if err := decoder.Decode(&decodeStore); err != nil {
+		return err
```
maxtaco: Is this an error out in the JSON decoding fails?

keybase/context_test.go
```diff
@@ -0,0 +1,33 @@
...
+	const message1 = "This is a test message\n"
+	const signature1 = `BEGIN KEYBASE SALTPACK DETACHED SIGNATURE.
+    kXR7VktZdyH7rvq v5wcIkPOwDJ1n11 M8RnkLKQGO2f3Bb fzCeMYz4S6oxLAy
+    Cco4N255JFQSlh7 IZiojdPCOssX5DX pEcVEdujw3EsDuI FOTpFB77NK4tqLr
+    Dgme7xtCaR4QRl2 hchPpr65lKLKSFy YVZcF2xUVN3gjpM vPFUMwg0JTBAG8x
+    Z. END KEYBASE SALTPACK DETACHED SIGNATURE.
+  `
+	reader := bytes.NewReader([]byte(message1))
+	err := ctx.Verify(reader, signature1)
+	assert.NoError(t, err)
```
maxtaco: is it worth testing: (1) sigs signed by other keys; and (2) bad sigs?

maxtaco:
Had a few questions...

keybase/config.go
gabriel: Whether the Auto value has been set by the user (via prompt)

keybase/config.go
gabriel: Yeah, the load fails, but the config is valid (falling back to default values). There is lots of testing around this... e.g. invalid config never prevents updates

keybase/context_test.go
gabriel: Yeah, have tests for that here: https://github.com/keybase/go-updater/blob/v2/saltpack_test.go

But worth adding for context since the interface is slightly different, and good to have overlap/layers of testing too... Will add now

maxtaco:
Thanks! 👍

PR #17, opened 2016-04-16
Utils for v2, including etag, digest, http save, unzip

gabriel:
@keybase/updater-hackers


util/digest_test.go
```diff
@@ -0,0 +1,30 @@
...
+func TestDigest(t *testing.T) {
+	data := []byte("test data\n")
+	path, err := WriteTempFile("TestDigest", data, 0644)
+	assert.NoError(t, err)
+	defer func() { _ = os.Remove(path) }()
+
+	err = CheckDigest("0c15e883dee85bb2f3540a47ec58f617a2547117f9096417ba5422268029f501", path, log)
+	assert.NoError(t, err)
+
+	err = CheckDigest("bad", path, log)
```
chrisnojima: should check the missing digest flow

util/etag.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"io"
+	"os"
+)
+
+// ComputeEtag returns etag for a file at path
+func ComputeEtag(path string) (string, error) {
```
chrisnojima: can't we just use the digest sha256 and not this md5 code?

util/unzip.go
```diff
@@ -0,0 +1,113 @@
...
+func UnzipOver(filename string, destinationPath string, log logging.Logger) error {
+	if !strings.HasSuffix(filename, ".zip") {
+		return fmt.Errorf("File has unsupported extension: %q", filename)
+	}
+
+	if destinationPath == "" {
+		return fmt.Errorf("Invalid destination %q", destinationPath)
+	}
+
+	if _, ferr := os.Stat(destinationPath); ferr == nil {
```
chrisnojima: so this'll remove it then try and unzip over it, is this dangerous? don't we want something like unzip into a temp, then do a name swap?

chrisnojima: i imagine a scenario like, i'm downloading a zip to replace /Applications/Keybase.app and i blow away the app, then unzip and that fails and now i have nothing

chrisnojima:
@jzila

util/etag.go
gabriel: Yeah the etag is for requesting from s3 with If-None-Match head which is to prevent the updater from redownloading ~50 megs over and over in case of the updater having trouble... we might want to disable this feature to simplify the updater though

util/unzip.go
gabriel: Yeah, the updater doesn't use this method to unzip over /Applications/Keybase, it uses this to unzip into a temp folder. The thing that moves into place is a rename which happens to be atomic on most platforms now.

util/digest_test.go
gabriel: added test

util/etag.go
chrisnojima: this is cause they write the etag themselves so they use md5? or do we control that also

util/etag.go
```diff
@@ -0,0 +1,28 @@
...
+	var result []byte
+	file, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer func() { _ = file.Close() }()
```
jzila: extraneous closure

util/http.go
```diff
@@ -40,6 +43,25 @@ func DiscardAndCloseBody(resp *http.Response) error {
...
+// SaveHTTPResponse saves an http.Response to path
+func SaveHTTPResponse(resp *http.Response, savePath string, mode os.FileMode, log logging.Logger) error {
+	if resp == nil {
+		return fmt.Errorf("No response")
+	}
+	file, err := os.OpenFile(savePath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, mode)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = file.Close() }()
```
jzila: extraneous closure

util/unzip.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// Closure to address file descriptors issue with all the deferred .Close() methods
+	extractAndWriteFile := func(f *zip.File) error {
+		rc, err := f.Open()
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err := rc.Close(); err != nil {
+				panic(err)
```
jzila: should this really be a panic?

jzila:
LGTM after nits.

util/etag.go
gabriel: Yeah this is a thing S3 supports, we don't make the etag, you have to use their generated etag

util/unzip.go
gabriel: Nope! Fixed

PR #21, opened 2016-04-20
Service, context, updater stubbing

gabriel:
This PR removes a lot of v1 code, updates some interfaces for v2.

- Context fixes
- Service starts up context (and starts checker)
- Update checker simplified for v2
- Remove singleflight
- Main updater code removed (stubbed out)


gabriel:
@keybase/updater-hackers

chrisnojima:
@keybase/react-hackers bump
@jzila

updater.go
```diff
@@ -3,833 +3,55 @@
...
-	switch err.(type) {
-	case CanceledError:
-		// Canceled errors aren't really errors
-		return keybase1.StatusOK(err.Error())
-	default:
-		return keybase1.FromError(err).Status()
-	}
+// Update performs an update
+func (u *Updater) Update(ctx Context) (*Update, error) {
+	// TODO
```
chrisnojima: ?

chrisnojima:
this method of merging into this brach and keeping the old code and overwriting it makes it logically hard to understand 'where we're at' in the process and what we'll end up with. unclear if there's something we can do at this point but its hard to feel confident in the PRs

chrisnojima:
@jzila can you take a look

keybase/context.go
```diff
@@ -23,48 +22,71 @@ var validCodeSigningKIDs = map[string]bool{
...
+}
+
 // UpdateOptions returns update options
 func (c *context) UpdateOptions() updater.UpdateOptions {
 	return c.config.updaterOptions()
 }
 
+// GetUpdateUI returns Update UI
+func (c *context) GetUpdateUI() (updater.UpdateUI, error) {
+	return c, nil
```
jzila: is this just to export the `UpdateUI` portions of this context?

keybase/context.go
```diff
@@ -23,48 +22,71 @@ var validCodeSigningKIDs = map[string]bool{
...
 	log logging.Logger
 }
 
-func newContext(cfg *config, log logging.Logger) *context {
-	ctx := context{}
-	ctx.config = cfg
-	ctx.log = log
+func newContext(upd *updater.Updater, cfg *config, log logging.Logger) *context {
+	ctx := context{
+		updater: upd,
```
jzila: I can't find where this is used

saltpack_test.go
```diff
@@ -58,28 +67,20 @@ func TestSaltpackVerifyBadValidIDs(t *testing.T) {
 
 	reader := bytes.NewReader([]byte(message1))
 	err := SaltpackVerifyDetached(reader, signature1, badCodeSigningKIDs, log)
+	assert.Error(t, err, "Should have failed verify")
 	t.Logf("Error: %s", err)
```
jzila: Do we want to log nil errors?

jzila:
Mostly looks good. Regarding the tests, it seems like you're using `testify/assert` and logging some of the errors, whereas the previous code would fatal out if a test condition failed. Do you want to use `testify/require` to replicate that?

LGTM after nits/questions.

keybase/context.go
gabriel: yeah, the context object implements the UpdateUI interface

keybase/context.go
gabriel: Yeah it's not used, will remove

saltpack_test.go
gabriel: Nope... did a pass where I changed some asserts to require

gabriel:
Yep, did a pass and changed some asserts to require.

gabriel:
@chrisnojima Yeah this PR was a bit messy cause I was removing code and re-adding stubs and a couple interfaces changed which affected different packages. I should have removed code in a separate PR and then done smaller PRs for the changes here, which became more obvious after I had made a bunch of changes in the same branch :(

PR #24, opened 2016-04-20
Keybase api (source) doesn't return status object

gabriel:
@maxtaco 

I'm ok with this.. if the HTTP response is not 200, then it's an error, the status object would be redundant


maxtaco:
We went back on forth on this but decided to leave it so it was compatible with the current clients, and before it would be less parsing work in Go if parsing into a fixed struct.

maxtaco:
👍

PR #25, opened 2016-04-20
Updater loop

gabriel:
This fills in some of the updater loop that is available.

The tests here are rudimentary, but we'll be adding more.

@keybase/updater-hackers

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -207,7 +207,7 @@ func (u *Updater) promptForUpdateAction(ctx Context, update Update, options Upda
 
 	u.log.Debugf("Update prompt response: %#v", updatePromptResponse)
 	if err := u.config.SetUpdateAuto(updatePromptResponse.AutoUpdate); err != nil {
-		u.log.Warningf("Error setting auto preference", err)
+		u.log.Warningf("Error setting auto preference: %s", err)
 	}
 
 	return updatePromptResponse.Action, nil
--- a/updater.go
+++ b/updater.go
@@ -3,7 +3,13 @@
 
 package updater
 
-import "github.com/keybase/go-logging"
+import (
+	"fmt"
+	"time"
+
+	"github.com/keybase/go-logging"
+	"github.com/keybase/go-updater/util"
+)
 
 // Version is the updater version
 const Version = "0.2.1"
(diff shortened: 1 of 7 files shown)


chrisnojima:
@keybase/react-hackers @jzila  bump

updater.go
```diff
@@ -50,8 +56,159 @@ func NewUpdater(source UpdateSource, config Config, log logging.Logger) *Updater
...
-	return nil, nil
+	options := ctx.UpdateOptions()
+	update, err := u.checkForUpdate(ctx, options)
+	if err != nil {
+		return nil, err
+	}
+	if update == nil {
+		// No update available
+		return nil, nil
+	}
```
jzila: can these two checks be combined into one:
```go
if update == nil {
    return nil, err
}`

File: updater.go [hunk 2]
```diff
@@ -50,8 +56,159 @@ func NewUpdater(source UpdateSource, config Config, log logging.Logger) *Updater
...
+		return nil, findErr
+	}
+	if update == nil {
+		return nil, nil
+	}
+
+	// Save InstallID if we received one
+	if update.InstallID != "" {
+		if err := u.config.SetInstallID(update.InstallID); err != nil {
+			u.log.Warningf("Error saving install ID: %s", err)
```
jzila: if there's a chance that the error returned can have different types, please use `%#v` to output the error.

updater.go
```diff
@@ -50,8 +56,159 @@ func NewUpdater(source UpdateSource, config Config, log logging.Logger) *Updater
...
+	autoUpdate := !autoSet
+	promptOptions := UpdatePromptOptions{AutoUpdate: autoUpdate}
+	updatePromptResponse, err := updateUI.UpdatePrompt(update, options, promptOptions)
+	if err != nil {
+		return UpdateActionError, err
+	}
+
+	u.log.Debugf("Update prompt response: %#v", updatePromptResponse)
+	if err := u.config.SetUpdateAuto(updatePromptResponse.AutoUpdate); err != nil {
+		u.log.Warningf("Error setting auto preference", err)
```
jzila: missing error format specifier?

updater_osx.go
```diff
@@ -89,6 +91,35 @@ func (u *Updater) openApplication(applicationPath string) error {
 }
 
 func (u *Updater) platformApplyUpdate(update Update, options UpdateOptions) error {
-	// TODO
+	localPath := update.Asset.LocalPath
+	destinationPath := options.DestinationPath
+	check := func(sourcePath string, destinationPath string) error {
```
jzila: why is this a closure?

updater_test.go
```diff
@@ -112,3 +116,49 @@ func newDefaultTestUpdateOptions() UpdateOptions {
...
+}
+
+func TestUpdater(t *testing.T) {
+	testServer := testServerForUpdateFile(t, testZipPath)
+	defer testServer.Close()
+
+	upr, err := newTestUpdaterWithServer(t, testServer)
+	assert.NoError(t, err)
+	update, err := upr.Update(testUpdateUI{newDefaultTestUpdateOptions()})
+	if assert.NoError(t, err) {
```
jzila: Hmm, I don't love this branched flow. I prefer to use the `require` package, which just terminates the test in the case that any assertion fails.

jzila:
LGTM after nits.

updater.go
gabriel: I kind of like being really verbose here and following the if err != nil pattern and then saying hey, no error but we don't have an update and that's ok...

If this seems crazy though I can change it to your suggestion.

updater.go
gabriel: Isn't `%#v` outputting go syntax? That seems to be a little awkward for logging given error interface is only `Error() string`?

updater.go
gabriel: Fixed

updater_osx.go
gabriel: Fixed

updater.go
jzila: yes but `%#v` outputs the _type_ of the error.

updater.go
jzila: naw, feel free to leave it.

updater_test.go
gabriel: Yes require.. was looking for that. Fixed.

updater.go
gabriel: I have the `Error() string` method returning the internal type in the description.

jzila:
👍

PR #29, opened 2016-04-25
More saltpack/verify tests

gabriel:
And more test coverage.

### DIFF
--- a/saltpack.go
+++ b/saltpack.go
@@ -29,32 +29,36 @@ func SaltpackVerifyDetachedFileAtPath(path string, signature string, validKIDs m
 	return nil
 }
 
+func checkSender(key saltpack.BasePublicKey, validKIDs map[string]bool, log logging.Logger) error {
+	if key == nil {
+		return fmt.Errorf("No key")
+	}
+	kid := key.ToKID()
+	if kid == nil {
+		return fmt.Errorf("No KID for key")
+	}
+	skid := hex.EncodeToString(kid)
+	log.Infof("Signed by %s", skid)
+	if validKIDs[skid] {
+		log.Debug("Valid KID")
+		return nil
+	}
+	return fmt.Errorf("Unknown signer KID: %s", skid)
+}
+
 // SaltpackVerifyDetached verifies a message signature
 func SaltpackVerifyDetached(reader io.Reader, signature string, validKIDs map[string]bool, log logging.Logger) error {
 	if reader == nil {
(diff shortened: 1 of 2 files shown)


gabriel:
@keybase/updater-hackers

saltpack_test.go
```diff
@@ -85,3 +90,31 @@ func TestSaltpackVerifyNoSignature(t *testing.T) {
...
+
+func TestSaltpackCheckNoKID(t *testing.T) {
+	err := checkSender(testSigningKey{kid: nil}, validCodeSigningKIDs, log)
+	require.Error(t, err)
+}
+
+func TestSaltpackVerifyNoFile(t *testing.T) {
+	err := SaltpackVerifyDetachedFileAtPath("/invalid", signature1, validCodeSigningKIDs, log)
+	require.Error(t, err)
+	t.Logf("Error: %s", err)
```
jzila: this line can never be reached

jzila: Maybe instead of logging the error, you can use [`require.EqualError`](https://godoc.org/github.com/jbert/testify/require#EqualError)

saltpack_test.go
```diff
@@ -85,3 +90,31 @@ func TestSaltpackVerifyNoSignature(t *testing.T) {
 	require.Error(t, err, "Should have failed verify")
 	t.Logf("Error: %s", err)
```
jzila: Maybe instead of logging the error, you can use [`require.EqualError`](https://godoc.org/github.com/jbert/testify/require#EqualError)

saltpack.go
```diff
@@ -29,32 +29,36 @@ func SaltpackVerifyDetachedFileAtPath(path string, signature string, validKIDs m
...
+	if kid == nil {
+		return fmt.Errorf("No KID for key")
+	}
+	skid := hex.EncodeToString(kid)
+	log.Infof("Signed by %s", skid)
+	if validKIDs[skid] {
+		log.Debug("Valid KID")
+		return nil
+	}
+	return fmt.Errorf("Unknown signer KID: %s", skid)
```
jzila: convention is error conditions first and `return nil` last, but not a big deal.

jzila:
lgtm after nits

saltpack.go
gabriel: Totally agree, flipped around

saltpack_test.go
gabriel: Fixed this everywhere in the file.. Also reading more testify docs :)

PR #38, opened 2016-04-26
App name changed to include Keybase (for process list, dock icon name etc)

gabriel:
@keybase/updater-hackers


gabriel:
@keybase/updater-hackers bump

jzila:
:+1:

PR #39, opened 2016-04-27
Report errors to the API

gabriel:
https://keybase.atlassian.net/browse/DESKTOP-902

I shorted UpdateError to just Error since this is our custom error for package updater.

### DIFF
--- a/error.go
+++ b/error.go
@@ -92,7 +92,3 @@ func restartErr(err error) Error {
 func configErr(err error) Error {
 	return NewError(ConfigError, err)
 }
-
-func errPtr(e Error) *Error {
-	return &e
-}
--- a/error.go
+++ b/error.go
@@ -5,62 +5,97 @@ package updater
 
 import "fmt"
 
-// UpdateErrorType is a unique short string denoting the error category
-type UpdateErrorType string
+// ErrorType is a unique short string denoting the error category
+type ErrorType string
 
 const (
-	// UpdateUnknownError is for if we had an unknown error
-	UpdateUnknownError UpdateErrorType = "unknown"
-	// UpdateCancelError is for if we canceled
-	UpdateCancelError UpdateErrorType = "cancel"
-	// UpdatePromptError is an UI prompt error
-	UpdatePromptError UpdateErrorType = "prompt"
-	// UpdateUnpackError is an error unpacking the asset
-	UpdateUnpackError UpdateErrorType = "unpack"
-	// UpdateCheckError is an error checking the asset
-	UpdateCheckError UpdateErrorType = "check"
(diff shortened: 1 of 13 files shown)


gabriel:
@keybase/updater-hackers

gabriel:
@keybase/updater-hackers bump ready

updater.go
```diff
@@ -57,11 +58,19 @@ func NewUpdater(source UpdateSource, config Config, log logging.Logger) *Updater
 }
 
 // Update checks, downloads and performs an update
-func (u *Updater) Update(ctx Context) (*Update, error) {
+func (u *Updater) Update(ctx Context) (*Update, *Error) {
```
maxtaco: why return a `*Error` rather than an `error` interface?

error.go
```diff
@@ -5,62 +5,97 @@ package updater
...
+func (e Error) Error() string {
 	if e.source == nil {
 		return fmt.Sprintf("Update Error (%s)", e.TypeString())
 	}
 	return fmt.Sprintf("Update Error (%s): %s", e.TypeString(), e.source.Error())
 }
 
-func cancelError(message string) UpdateError {
-	return NewUpdateError(UpdateCancelError, fmt.Errorf(message))
+func cancelErrPtr(err error) *Error {
```
maxtaco: this technique makes me nervous because of:

```go
var e1 *Error;
var e2 error;
e2 = e1;
if e2 != nil {
    fmt.Printf("WTF!\n")
}
```

maxtaco:
Big question surrounding the use of an `*Error`, which I worry could be buggy.  Otherwise looks good.

updater.go
gabriel: Yeah, sure, exposing custom error types in a public interface isn't common... and totally unnecessary here

maxtaco: Cool, this will save you from ever needing an `*Error` type, which can be buggy because of the crazy `nil` v `nil` semantics in Go.

gabriel:
I changed around the error helper functions to not return pointer.. instead using an errPtr function to change Error to pointer for return... basically I'm trying to avoid this:

```go
uerr := cancelErr(err)
return update, &uerr
```

by doing this:

```go
return update, errPtr(cancelErr(err))
```

since you can't do `return update, &(cancelError(err))` in go

Is the better?

gabriel:
Actually, I think I understand... the golang way is the use the error interface throughout and then use a type switch to get a the specific error when needed? 

If in reporting, for some reason the error is not out updater.Error type, then the error_type in reporting will be "unknown".

jzila:
@gabriel yeah, error interface + type switch is the way to go.

PR #40, opened 2016-04-27
Report action to API

gabriel:
The gocyclo score for updater.Update() got too high so had to reduce it a little.

@keybase/updater-hackers

### DIFF
--- a/keybase/report.go
+++ b/keybase/report.go
@@ -15,19 +15,27 @@ import (
 )
 
 // ReportError notifies the API server of a client updater error
-func (c context) ReportError(updateErr updater.Error, options updater.UpdateOptions) {
-	if err := c.reportError(updateErr, options, defaultEndpoints.err, time.Minute); err != nil {
+func (c context) ReportError(err error, options updater.UpdateOptions) {
+	if err := c.reportError(err, options, defaultEndpoints.err, time.Minute); err != nil {
 		c.log.Warningf("Error notifying about an error: %s", err)
 	}
 }
 
-func (c context) reportError(updateErr updater.Error, options updater.UpdateOptions, uri string, timeout time.Duration) error {
+func (c context) reportError(err error, options updater.UpdateOptions, uri string, timeout time.Duration) error {
+	var errorType string
+	switch uerr := err.(type) {
+	case updater.Error:
+		errorType = uerr.TypeString()
+	default:
+		errorType = string(updater.UnknownError)
+	}
+
 	data := url.Values{}
 	data.Add("install_id", options.InstallID)
 	data.Add("version", options.Version)
 	data.Add("upd_version", options.UpdaterVersion)
-	data.Add("error_type", updateErr.TypeString())
-	data.Add("description", updateErr.Error())
+	data.Add("error_type", errorType)
+	data.Add("description", err.Error())
 	return c.report(data, uri, timeout)
 }
 
--- a/keybase/report.go
+++ b/keybase/report.go
@@ -1,0 +1,50 @@
+// Copyright 2015 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package keybase
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/keybase/go-updater"
+	"github.com/keybase/go-updater/util"
+)
+
+// ReportError notifies the API server of a client updater error
+func (c context) ReportError(updateErr updater.Error, options updater.UpdateOptions) {
+	if err := c.reportError(updateErr, options, defaultEndpoints.err, time.Minute); err != nil {
+		c.log.Warningf("Error notifying about an error: %s", err)
+	}
+}
+
+func (c context) reportError(updateErr updater.Error, options updater.UpdateOptions, uri string, timeout time.Duration) error {
+	data := url.Values{}
+	data.Add("install_id", options.InstallID)
+	data.Add("version", options.Version)
+	data.Add("upd_version", options.UpdaterVersion)
+	data.Add("error_type", updateErr.TypeString())
+	data.Add("description", updateErr.Error())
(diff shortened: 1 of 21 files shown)


gabriel:
@keybase/updater-hackers bump

updater.go
```diff
@@ -89,9 +90,12 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, *Error) {
 		return update, promptErrPtr(err)
 	}
 	switch updateAction {
-	case UpdateActionApply, UpdateActionAuto:
-		// Continue
+	case UpdateActionApply:
+		ctx.ReportAction(UpdateActionApply, options)
+	case UpdateActionAuto:
+		ctx.ReportAction(UpdateActionAuto, options)
```
maxtaco: Is this when they chose to apply the update, and also have clicked the "make next update auto" button?

keybase/report.go
```diff
@@ -31,6 +31,22 @@ func (c context) reportError(updateErr updater.Error, options updater.UpdateOpti
...
+	}
+}
+
+func (c context) reportAction(action updater.UpdateAction, options updater.UpdateOptions, uri string, timeout time.Duration) error {
+	data := url.Values{}
+	data.Add("install_id", options.InstallID)
+	data.Add("version", options.Version)
+	data.Add("upd_version", options.UpdaterVersion)
+	data.Add("action", action.String())
+	return c.report(data, uri, timeout)
```
jzila: should this maybe happen asynchronously?

jzila: you can return a channel that'll get closed when the task is done.

updater.go
```diff
@@ -126,23 +130,31 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, *Error) {
...
-	if err := ctx.BeforeApply(*update); err != nil {
-		return update, applyErrPtr(err)
+	if err := u.apply(ctx, *update, options); err != nil {
+		return update, err
 	}
 
-	if err := u.platformApplyUpdate(*update, options); err != nil {
-		return update, applyErrPtr(err)
+	if err := ctx.Restart(); err != nil {
+		return update, restartErrPtr(err)
```
jzila: I agree with Max on the other PR, these should really be `error` interface.

updater_test.go
```diff
@@ -154,13 +166,20 @@ func testServerForError(t *testing.T, err error) *httptest.Server {
 	}))
 }
 
+func testServerNotFound(t *testing.T) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jzila: TIL `httptest`. Nice!

jzila:
A few nits. Looks good otherwise.

updater.go
gabriel: auto is if the update was applied automatically without user prompt...

so the first time the action will be apply, and if they checked automatic, then next time it would be auto

keybase/report.go
gabriel: there is a minute timeout on all reporting requests.. but I think yeah.. the reporting actions shouldn't slow down an update... I hesitate to introduce async behavior here only because it's harder to reason about it

updater.go
maxtaco: OK. The way it's currently implemented, if you send `apply` then the auto bit is set to `false`, and if you send `auto`, the auto bit is set to `true`:

https://github.com/keybase/keybase/pull/647/files#diff-09c19ea087354ad87b44929d4a676ff1R224

gabriel: is it important to distinguish between... this update was applied without user being prompted... in which case my reporting is correct... if the goal is to report the "auto" setting then we need to change this request to not mix an action with state?

So params would be?
```
action: "apply", "snooze", "cancel", "auto"
auto_apply: 1/0
```

If apply is selected and user doesn't enable auto: `action=apply&auto_apply=0`
If apply is selected and user turns on auto apply: `action=apply&auto_apply=1`
If there was no prompt, user turns on auto apply: `action=auto&auto_apply=1`

If we ever see auto with auto_apply=0 then that would indicate a problem.

updater.go
gabriel: Yeah I suppose the golang way is to use the error interface everywhere and to use a type switch when you need to check for specific error type?

It always seemed "code smell" to be checking types and casting but I suppose in go since you don't have inheritance using type switches isn't a bad thing?

maxtaco: Yes, I think you're exactly right.  I too resisted type-casts, since in C++, it's a sign of weakness. But I think Go wants to you to do it this way. @jzila do you concur?

updater.go
maxtaco: I'm fine with this change, though I would have predicted that you wouldn't like a 2x2 of bools in which one quadrant is undefined behavior :)

maxtaco: Ok, changed the pseudocode and confluence doc to match.  BTW, I never implemented `cancel` here.

maxtaco: Ok, changes are up here: https://github.com/keybase/keybase/pull/647/commits/1861e3d0fd9f13dbda7f70037ac27ad6efc5e877

keybase/report.go
jzila: With channels, async behavior isn't that much more difficult to reason about. Especially when it's effectively just a promise: return a channel, and then the caller blocks at will until the channel is closed.

updater.go
jzila: 100% in support of type switch or type assertion checking of errors. It's definitely the golang way, and it's the standard way of handling not just errors but any time you need to resolve an interface.

updater.go
gabriel: Should I remove reporting of cancel? It's probably not useful. Only happens in case of prompt timeout.

gabriel: I can report it and the server can ignore it if it wants to?

gabriel:
Gonna add async reporting and auto_update in separate PRs.

PR #41, opened 2016-04-28
Move command.go to package

gabriel:
- Changing RunCommand to command.Exec
- Returns Result object with bytes.Buffer (not combined output)

This allows command execution to be more precise and explicit.

There was really good test coverage on this.


gabriel:
@keybase/updater-hackers

gabriel:
I think coveralls is confused :(

jzila:
:+1: 

I think coveralls is correct, but is only tracking coverage on `updater` package. Your `command` coverage was higher than the `updater` coverage, so by moving it out, the `updater` coverage was reduced (or vice versa, not quite sure)

PR #43, opened 2016-04-29
App restart (darwin)

gabriel:
This uses ps to find the pid of an app, since app bundles use an executable
like Test.app/Contents/MacOS/Test. It will SIGTERM this process (and SIGKILL
after a second), and then call open.

It doesn't fail to call open if the kill failed.

Bundled is a Test.app that is used in the test.


process/process_darwin.go
```diff
@@ -0,0 +1,134 @@
...
+	if process == nil {
+		return fmt.Errorf("No process found with pid %d", pid)
+	}
+
+	log.Debugf("Terminating: %#v", process)
+	err = process.Signal(syscall.SIGTERM)
+	if err != nil {
+		log.Warningf("Error sending terminate: %s", err)
+	}
+	time.Sleep(time.Second)
```
jzila: since this is called in a loop, you may wish to do this sleep-then-kill part in a goroutine.

gabriel:
This changed a little bit... instead of using ps we're using go-ps library.

New PR here: https://github.com/keybase/go-updater/pull/48

The open and term are the same and the find pid methods are much simpler (since parsing not required).

process/process_darwin.go
gabriel: it would be very rare if there were more than 1 process/app running for a particular app bundle... so am ok with it being in serial

PR #44, opened 2016-04-29
Testing for update sources (local, remote)

gabriel:
These are update sources useful for testing:

- local.go is for pulling an update locally, which is useful for testing and in development
- remote.go is for pulling an update from a generic remote source like S3, also useful for testing.

I'm moving them into a sources package since they don't need to be part of the "core" library.


gabriel:
@keybase/updater-hackers

jzila:
LGTM

PR #46, opened 2016-05-02
Don't pre-download update asset (before prompting)

gabriel:
This PR only removes code.

This was inherited from v1 but it makes things a bit more complex for not a lot of gain,
and arguably isn't typical behavior for an updater.


gabriel:
@keybase/updater-hackers

jzila:
👍

PR #47, opened 2016-05-02
Send auto_update config for update and action report

gabriel:
As discussed in this PR here: https://github.com/keybase/go-updater/pull/40#discussion_r61696909
Is also in design doc.

### DIFF
--- a/keybase/report.go
+++ b/keybase/report.go
@@ -52,6 +52,8 @@ func (c context) reportAction(action updater.UpdateAction, options updater.Updat
 	data.Add("version", options.Version)
 	data.Add("upd_version", options.UpdaterVersion)
 	data.Add("action", action.String())
+	autoUpdate, _ := c.config.GetUpdateAuto()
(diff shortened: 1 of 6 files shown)


gabriel:
@keybase/updater-hackers

keybase/report.go
```diff
@@ -52,6 +52,8 @@ func (c context) reportAction(action updater.UpdateAction, options updater.Updat
 	data.Add("version", options.Version)
 	data.Add("upd_version", options.UpdaterVersion)
 	data.Add("action", action.String())
+	autoUpdate, _ := c.config.GetUpdateAuto()
```
maxtaco: If we've made the change on this iteration through the loop, is it reflected in this getter?

gabriel: yeah, the autoUpdate config is set/saved immediately after the prompt returns... and the action is reported after this... so we should be all good... I need to add more tests for this though

gabriel: Added test

jzila:
lgtm

PR #48, opened 2016-05-02
App restart

gabriel:
Bundled in test dir is a Test.app that is used in a darwin to test restarting
and app bundle.


gabriel:
I'm investigating Windows, but will deal with that in a separate PR.

gabriel:
@keybase/updater-hackers

keybase/platform_windows.go
```diff
@@ -43,3 +43,7 @@ func (c context) UpdatePrompt(update updater.Update, options updater.UpdateOptio
 func (c context) PausedPrompt() error {
 	return fmt.Errorf("Unsupported")
 }
+
+func (c context) Restart() error {
```
chrisnojima: we should have a todo or throw an error here right, same w/ linux

process/process_darwin.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/keybase/go-updater/command"
+)
+
+// RestartAppDarwin restarts an app. We will still call open if the kill fails.
+func RestartAppDarwin(appPath string, log logging.Logger) error {
+	if appPath == "" {
+		return fmt.Errorf("No app path to restart")
+	}
+	procName := filepath.Join(appPath, "Contents/MacOS/")
+	TerminateAll(procName, log)
```
chrisnojima: I thought we were going to send a ctl stop to the app first, then do this

gabriel: yeah I was temporarily holding off on this, since it seems Windows does support SIGTERM in which cause we can rely on terminating on pids

gabriel: after doing windows testing, it doesn't support SIGTERM (only in some weird situations it might)... will add ctl stop.

gabriel: ok talking to steve.. we can wait on doing ctl stop until we implement windows, which will come in a separate PR.
It's not required for darwin/linux since sending SIGTERM is all thats needed on those platforms.

chrisnojima: Why is sig term better than calling ctl stop

gabriel:
bump

jzila:
I thought I already approved this PR. LGTM

PR #50, opened 2016-05-03
Report (notify) success

gabriel:

Will add request_id to all reporting in a different PR

### DIFF
--- a/keybase/report_test.go
+++ b/keybase/report_test.go
@@ -77,3 +77,18 @@ func TestReportActionEmpty(t *testing.T) {
 	err := ctx.reportAction("", updater.UpdateOptions{}, server.URL, 5*time.Millisecond)
 	assert.NoError(t, err)
 }
+
+func TestReportSuccess(t *testing.T) {
+	server := newServer("{}")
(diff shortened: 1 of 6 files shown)


keybase/report_test.go
```diff
@@ -77,3 +77,18 @@ func TestReportActionEmpty(t *testing.T) {
...
+func TestReportSuccess(t *testing.T) {
+	server := newServer("{}")
+	defer server.Close()
+
+	tmp := defaultEndpoints.success
+	defaultEndpoints.success = server.URL
+	defer func() { defaultEndpoints.success = tmp }()
+
+	ctx := testContext(t)
+	ctx.ReportSuccess(testOptions)
```
jzila: Technically this improves coverage, but we can't validate the result of this exported call, and we _do_ validate its effective result below. So I'm not sure it's useful.

jzila:
Just one nit, otherwise LGTM.

keybase/report_test.go
gabriel: yeah it's unnecessary.. removed

PR #51, opened 2016-05-03
Adding cert and TLS config for api.keybase.io

gabriel:
https://keybase.atlassian.net/browse/DESKTOP-925


patrickxb:
Seems kind of a shame that we aren't using `libkb/client.go` for this...

For example, what about proxies (which `libkb/client.go` supports)?  

Another thing `libkb/client.go` has support for is Tor mode.

maxtaco:
It's a goal of the updated to have minimal dependencies, so it's probably the right call...

patrickxb:
Ok, well then should it copy the proxy support from `libkb/client.go` too?  Seems like we would want updates to work with proxies...

maxtaco:
Hmm, that's a good point....

gabriel:
I could read proxyCAs (or proxy_ca_certs) key from config.json?

Honestly, libkb is too big of a monster to include as a dependency...

Also I noticed these config key names are inconsistent: https://github.com/keybase/client/pull/2768

gabriel:
Can I do proxy support as separate ticket? Created here: https://keybase.atlassian.net/browse/DESKTOP-744

maxtaco:
Sure.  You might need it across all HTTP actions. 

👍 for me!

jzila:
🚢

PR #53, opened 2016-05-03
Request ID

gabriel:
Implements RequestID.

Also remove InstallID from options since its part of update response and config.

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -36,8 +36,8 @@ type Context interface {
 	BeforeApply(update Update) error
 	AfterApply(update Update) error
 	Restart() error
-	ReportError(err error, options UpdateOptions)
-	ReportAction(action UpdateAction, options UpdateOptions)
+	ReportError(err error, update *Update, options UpdateOptions)
+	ReportAction(action UpdateAction, update *Update, options UpdateOptions)
 }
 
 // Config defines configuration for the Updater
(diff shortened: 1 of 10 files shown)


gabriel:

updater.go
```diff
@@ -36,8 +36,8 @@ type Context interface {
 	BeforeApply(update Update) error
 	AfterApply(update Update) error
 	Restart() error
-	ReportError(err error, options UpdateOptions)
-	ReportAction(action UpdateAction, options UpdateOptions)
+	ReportError(err error, update *Update, options UpdateOptions)
+	ReportAction(action UpdateAction, update *Update, options UpdateOptions)
```
jzila: Is this change to the exported interface only to enable test validation?

jzila:
Just one pretty big question.

updater.go
gabriel: Nope, it's to give the reporting access to the update, which includes the RequestID returned by the find update request.

maxtaco:
I'm late to the game, but i didn't see how the `requestID` comes back from the `update.json` call and then is threaded through all of the subsequent calls.  Also, I didn't see how `requestID` changes every time through the loop, while `InstallID` stays fixed.

gabriel:
request_id gets passed to reporting stuff via update object... new requestId gets returned by update.json request in each loop I assume?

install_id same thing, except it doesn't change... the installId returned from update.json is either a new one or echoing back the one passed in?

maxtaco:
this makes sense, thanks!

PR #62, opened 2016-05-09
Exit for restart

gabriel:
Exiting after update check allows the updater to restart.

From the comment above the exit:
```go
// If we received an update from the check let's exit, so the watchdog
// process (e.g. launchd on darwin) can restart us, no matter what, even if
// there was an error, and even if the update was or wasn't applied.
// There is no difference between doing another update check in a loop after
// delay and restarting the service.
```

Also temporarily setting test channel, bumping version and adding some logging.

### DIFF
--- a/update_checker.go
+++ b/update_checker.go
@@ -38,9 +38,11 @@ func (u *UpdateChecker) check() error {
 	u.count++
 	update, err := u.updater.Update(u.ctx)
 	if update != nil {
-		// If we received an update let's exit, so the watchdog process can restart
-		// us, no matter what, even if there was an error, and even if the update
-		//  was or wasn't applied.
+		// If we received an update from the check let's exit, so the watchdog
+		// process (e.g. launchd of darwin) can restart us, no matter what, even if
+		// there was an error, and even if the update was or wasn't applied.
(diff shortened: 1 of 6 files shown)


gabriel:
@keybase/updater-hackers

gabriel:
Gonna merge this in prematurely, cause this allows us to start dogfooding

update_checker.go
```diff
@@ -21,7 +22,7 @@ type UpdateChecker struct {
 
 // NewUpdateChecker creates an update checker
 func NewUpdateChecker(updater *Updater, ctx Context, log logging.Logger) UpdateChecker {
-	return newUpdateChecker(updater, ctx, log, DefaultTickDuration())
+	return newUpdateChecker(updater, ctx, log, time.Hour)
```
jzila: shouldn't this be a constant somewhere?

jzila:
LGTM, one nit but it's not critical.

update_checker.go
gabriel: yup.. in this PR, the ticker stuff got better (default is a const): https://github.com/keybase/go-updater/pull/65

PR #63, opened 2016-05-09
Util tweaks

gabriel:
- Move rand util stuff into rand.go
- RandString changed to RandomID and uses 256 bits
- Move string util into strings.go

### DIFF
--- a/util/rand_test.go
+++ b/util/rand_test.go
@@ -9,12 +9,12 @@ import (
 )
 
 func TestRandString(t *testing.T) {
-	s, err := RandomID("prefix.")
+	s, err := RandomID("prefix=")
 	t.Logf("Rand string: %s", s)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if !strings.HasPrefix(s, "prefix.") {
+	if !strings.HasPrefix(s, "prefix=") {
 		t.Errorf("Invalid prefix: %s", s)
 	}
 	if len(s)-len("prefix.") != 52 {
--- a/util/rand_test.go
+++ b/util/rand_test.go
@@ -1,0 +1,23 @@
+// Copyright 2015 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package util
+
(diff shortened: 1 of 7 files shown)


gabriel:
@keybase/updater-hackers

util/rand_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestRandString(t *testing.T) {
+	s, err := RandomID("prefix.")
+	t.Logf("Rand string: %s", s)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !strings.HasPrefix(s, "prefix.") {
+		t.Errorf("Invalid prefix: %s", s)
+	}
+	if len(s)-len("prefix.") != 52 {
```
chrisnojima: does this actually work if the random number encodes with padding (which then gets removed)?

gabriel: Good catch, yeah the padding remove should only happen from the returned base32 string. Fixed with test.

util/rand_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestRandString(t *testing.T) {
+	s, err := RandomID("prefix=")
+	t.Logf("Rand string: %s", s)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !strings.HasPrefix(s, "prefix=") {
+		t.Errorf("Invalid prefix: %s", s)
+	}
+	if len(s)-len("prefix.") != 52 {
```
chrisnojima: even with the above fix won't this fail sometimes? sometimes the base32 encoding will have padding / won't (or is that not true). It seems like if it does sometimes and doesn't sometimes, then the length of the random string (ignoring prefix) will be variable length (cause we strip it sometimes) so it won't always be 52 long. like sometimes it'll be 51 and still be working fine

gabriel: ah.. the random id is always 256 bits (as suggested by https://gist.github.com/tqbf/be58d2d39690c3b366ad)... this means the padding will always be the same

PR #64, opened 2016-05-10
Command line update check

gabriel:
Add a "check" arg, so you can perform an update check manually from command line.

This has a super minimal impact so I think safe to add in the service exectuable.

@keybase/updater-hackers

### DIFF
--- a/service/main.go
+++ b/service/main.go
@@ -6,7 +6,7 @@ package main
 import (
 	"flag"
 	"fmt"
-	"os"
+	"log"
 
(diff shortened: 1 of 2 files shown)


gabriel:
e.g. for example on OS X this would be:

`/Applications/Keybase.app/Contents/SharedSupport/bin/updater check`

gabriel:
@keybase/updater-hackers

jzila:
lgtm

PR #67, opened 2016-05-13
Windows test fixes

gabriel:
- Replacing use of shell scripts with a test go program (which is cross platform)
- Extend some timeouts which can cause flakiness on slow windows machine
- Don't SIGTERM on windows
- Change some EqualError assertions that don't apply on Windows
- --Skip-- Fix symlink tests on Windows
- Skip file permission tests on Windows
- Fix path bugs on Windows


gabriel:
@keybase/updater-hackers

util/file.go
```diff
@@ -131,6 +134,7 @@ func openTempFile(prefix string, suffix string, mode os.FileMode) (string, *os.F
 
 // FileExists returns whether the given file or directory exists or not
 func FileExists(path string) (bool, error) {
+	fmt.Printf("%s\n", path)
```
chrisnojima: 🔥

util/file.go
```diff
@@ -289,3 +293,33 @@ func ReadFile(path string) ([]byte, error) {
...
+
+// URLStringForPath returns an URL as string with file scheme for path.
+// For example,
+//     /usr/local/go/bin => file:///usr/local/go/bin
+//     C:\Go\bin => file:///C:/Go/bin
+func URLStringForPath(path string) string {
+	switch runtime.GOOS {
+	case "windows":
+		// Include leading slash on windows
+		return fmt.Sprintf("%s:///%s", fileScheme, filepath.ToSlash(path))
```
chrisnojima: doesn't this need to be escaped? paths with spaces and etc

util/file_test.go
```diff
@@ -25,8 +27,11 @@ func TestNewFile(t *testing.T) {
 
 	fileInfo, err := os.Stat(filename)
 	assert.NoError(t, err)
-	assert.True(t, 0600 == fileInfo.Mode().Perm())
 	assert.False(t, fileInfo.IsDir())
+
+	if runtime.GOOS != "windows" {
+		assert.True(t, 0600 == fileInfo.Mode().Perm())
```
chrisnojima: does windows have no concept of this or is the value different. if the latter we should check it

util/file_test.go
```diff
@@ -95,14 +106,17 @@ func TestIsDirReal(t *testing.T) {
...
 
-	symLinkPath := TempPath("", "TestIsDirReal")
-	err = os.Symlink(os.TempDir(), symLinkPath)
-	defer RemoveFileAtPath(symLinkPath)
-	assert.NoError(t, err)
-	ok, err = IsDirReal(symLinkPath)
-	assert.Error(t, err)
-	assert.Equal(t, "Path is a symlink", err.Error())
-	assert.False(t, ok)
+	// Windows requires privileges to create symbolic links
```
chrisnojima: is this true? in our repo we use mklink /j which (i don't think) requires elevated priveledges

util/file_test.go
```diff
@@ -267,5 +281,26 @@ func TestReadFile(t *testing.T) {
 	assert.Equal(t, dataIn, dataOut)
 
 	_, err = ReadFile("/invalid")
-	assert.EqualError(t, err, "open /invalid: no such file or directory")
+	assert.Error(t, err)
+	require.True(t, strings.HasPrefix(err.Error(), "open /invalid: "))
+}
+
+func TestURLStringForPath(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
chrisnojima: we need to test more complicated paths, like those with weird chars / etc in them

chrisnojima:
@jzila 👀

util/file_test.go
gabriel: yeah windows doesn't have this

util/file.go
gabriel: Added path escaping and tests

util/file_test.go
gabriel: Fixed the test to do mklink /j on windows

gabriel:
bump

PR #68, opened 2016-05-14
Update go-ps library

gabriel:
This fixes the updater to use this recent fixes in the go-ps library.

The go-ps library had inconsistent values for process Executable() across platforms. The go-ps change vendored here makes the Executable() be the process filename on all platforms, and adds a Path() method for the full path on supported platforms.


gabriel:
@keybase/updater-hackers

jzila:
lgtm

PR #69, opened 2016-05-19
Don't re-save install id if unchanged

gabriel:
Also changes some log statements to be a little clearer/less misleading.


gabriel:
@keybase/updater-hackers

jzila:
lgtm

PR #70, opened 2016-05-19
Remove unnecessary file processing

gabriel:
We were chowning and touching files after unziping, but the unzip
method doesn't preserve (already uses current user and time).

Also this processing was happening after the app was moved into
place which though it worked, doesn't feel like a good idea.. cause
you know, don't fuck with Gatekeeper :)


gabriel:
@keybase/updater-hackers

gabriel:
bump

jzila:
lgtm

PR #71, opened 2016-05-19
Use a temp dir to move app before applying update (instead of removing)

gabriel:
This PR creates a temp dir at start of the update which is passed to apply.
The apply moves existing app into this temp dir.
At the end of update it is removed (relies on defer).


gabriel:
After more research this isn't really a problem and causes the process path to move

gabriel:
@keybase/updater-hackers

chrisnojima:
What problem does this solve? Is this updated in the confluence doc?

gabriel:
Put this section in the confluence doc:

> While it is theoretically ok to remove an executable (or bundle) while that process is running (inodes are not removed until process file handles are closed on unix style systems), in order to minimize any potential flakeiness or bugs where this assumption might not be unknowingly held (for Gatekeeper or launchd on OS X, for example). It's a good idea to move the old app bundle to a temporary directory, before applying the update and restarting (and then doing removal after the process has ended).
This currently only applies to OS X, but for Windows removing a file while a process is running is not necessarily safe (AFAIK).

> The other reason this is better is that (on OS X) the proc_listpids function (used by go-ps library) may not return processes for things running that no longer exist on disk, since the proc_pidpath doesn't exist anymore. I've investigating this more though.

> All in all, I think this is a safer approach that reduces potential unforseen consequences of removing a running process from disk.

This all came about from doing a bunch of testing (brute force applying updates, re-applying updates with same version and verifying logs, etc).

cjb:
The approach sounds good to me, but please get a golang expert to review the code.

@keybase/updater-hackers

jzila:
lgtm

PR #74, opened 2016-05-23
Check and report status of restart

gabriel:
Also includes some more process util methods.


gabriel:
@keybase/updater-hackers

keybase/platform_darwin.go
```diff
@@ -105,17 +106,33 @@ func (c context) Restart() error {
 		return fmt.Errorf("No destination path for restart")
 	}
 
-	procName := filepath.Join(appPath, "Contents/MacOS/")
-	process.TerminateAll(procName, time.Second, c.log)
+	serviceProcPath := "Keybase.app/Contents/SharedSupport/bin/keybase"
```
jzila: can these be constants?

process/process.go
```diff
@@ -17,41 +17,109 @@ import (
 
 type processesFn func() ([]ps.Process, error)
 
-func findPIDsWithFn(fn processesFn, prefix string, log logging.Logger) ([]int, error) {
-	processes, err := fn()
+type matchFn func(ps.Process) bool
+
+func matchPath(p ps.Process, match string, log logging.Logger) bool {
```
jzila: whoa I think you just found a github syntax highlighting bug!

jzila: also you might want to make the name clearer given that it returns a bool, say `matchesPath`, or `doesMatchPath`

process/process.go
```diff
@@ -17,41 +17,109 @@ import (
...
+	}
+	return strings.Contains(path, match)
+}
+
+// FindProcesses returns processes containing string matching process path
+func FindProcesses(match string, wait time.Duration, delay time.Duration, log logging.Logger) ([]ps.Process, error) {
+	log.Infof("Finding process %q (max wait %s)", match, wait)
+	matchPath := func(p ps.Process) bool { return containsPath(p, match, log) }
+	start := time.Now()
+	for i := 0; time.Since(start) < wait || i == 0; i++ {
```
jzila: I think this would be clearer as:
```go
for {
    ...
    if time.Since(start) >= wait {
        break
    }
}

File: process/process.go [hunk 4]
```diff
@@ -17,41 +17,109 @@ import (
...
+		return nil, err
+	}
+	if len(procs) == 0 {
+		return nil, nil
+	}
+	return procs[0], nil
+}
+
+// Ignore deadcode warning
+var _ = findProcessWithPID
```
jzila: If the linter is causing us to do things like this, we should re-evaluate that linter rule.

process/process.go
gabriel: Refactored the match functions into a Matcher with match types (PathEqual/PathContains/...)

process/process.go
gabriel: So this is a useful function but is currently only used by tests... I don't want to nuke it though and don't want to move it into test files. Basically am ok with a deadcode ignore here.... I made the comment explain this more.

jzila: In that case, I'd rather put it in test files for now, and move it once regular code uses it.

process/process.go
```diff
@@ -17,41 +17,109 @@ import (
...
+
+// findProcessesWithFn finds processes using match function.
+// If max is != 0, then we will return that max number of processes.
+func findProcessesWithFn(processesFn processesFn, matchFn matchFn, max int) ([]ps.Process, error) {
```
jzila: I'd prefer if the identifier and type were named something different.

gabriel: sure

gabriel:
ok I think it's finally ready...

process/process_test.go
```diff
@@ -72,3 +75,41 @@ func TestTerminateAllFn(t *testing.T) {
...
+}
+
+func TestFindProcessWait(t *testing.T) {
+	path := "/bin/echo"
+	go func() {
+		time.Sleep(10 * time.Millisecond)
+		cmd := exec.Command(path, "hi")
+		err := cmd.Start()
+		require.NoError(t, err)
+	}()
```
jzila: I think this is asking for flakiness :(

jzila: maybe if it accepts a channel it can be tested with github.com/jonboulle/clockwork

jzila:
lgtm after nits

process/process_test.go
gabriel: Yeah, the issue is that it is actually listing processes on the system, so it's hard or impossible to use clockwork without also mocking the process listing part, which is an option (and happens in other tests), but I was able to make this not flakey by moving the "not running" check before the sleep, so I think that's probably better... if you feel really strongly though... I guess all these millisecond sleeps do add up... the process tests currently run in ~300ms total so I think still ok to avoid clockwork..

jzila: Yeah I'm less worried about the time added and more worried about potential flakiness. I guess we'll see :)

PR #75, opened 2016-05-24
Send force param to API (update.json)

gabriel:
The API should return NeedUpdate to true on response and include a requestId when passing force=1 to update.json endpoint.


gabriel:
cc @maxtaco @cjb

maxtaco:
👍   I have a PR for the server side here: https://github.com/keybase/keybase/pull/680/files

cjb:
@gabriel please merge, server side is in (though seems broken right now)

PR #78, opened 2016-05-25
Watchdog

gabriel:
From the README:

> The watchdog is in charge of ensuring that a list of programs are always running.
> The watchdog will start these programs (it will be the parent of those processes).
> If the watchdog is terminated, the monitored programs will also be terminated.

> When the watchdog starts up, it will kill any programs that it will be monitoring.
> Then it will monitor a list of programs and will restart them if they exit.
> Programs can be configured to always run, or to stop when receiving a particular
exit status.


gabriel:
@zanderz I will have a PR for the client (ctl watchdog) that uses this package very soon, tomorrow morningish

gabriel:
cc @keybase/updater-hackers

watchdog/watchdog_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/keybase/go-updater/process"
+	"github.com/keybase/go-updater/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var testLog = &logging.Logger{Module: "test"}
+
+func TestWatch(t *testing.T) {
+	procProgram1 := procProgram(t, "testWatch1", 60)
```
chrisnojima: where's the test for it killing an existing program on start

watchdog/watchdog_test.go
```diff
@@ -0,0 +1,76 @@
...
+		_ = cmd.Process.Kill()
+	}
+	_ = os.Remove(procPath)
+}
+
+// procProgram returns a testable unique program at a temporary location
+func procProgram(t *testing.T, name string, delaySeconds int) Program {
+	// Copy sleep executable to tmp
+	procPath := filepath.Join(os.TempDir(), name)
+	err := util.CopyFile("/bin/sleep", procPath, testLog)
```
zanderz: This won't work on Windows. Maybe use calc.exe or somesuch?

zanderz:
There are a couple of things the old watchdog had that this doesn't, just checking that we're getting rid of them on purpose.
1. Old one only restarted the service 10 times, to avoid thrashing in case of a massive failure
2. A bunch of logic was added for `ctl restart` so that the watchdog wouldn't start an extra copy in that case.

gabriel:
(1) I thought it was better to remove the 10 failures max strategy and follow the same launchd strategry as launchd (on OS X), which is very simple.. to restart a program after 10 seconds if it errors.. it will try forever and has no backoff strategy. 

The thing I worried about with the 10 failures max strategy was that it was over the total lifetime of the process... so 10 failures over the span of a few days could cause it to stop monitoring. And maybe if perhaps something on the server was triggering a crash, we could fix on the server and have clients recover despite more than 10 errors.

(2) The watchdog will lookup and stop any existing services it will be monitoring first. This prevents multiple services from running... Since this needs to happen for all services, not just for client, I do this by looking up in the process table. The downside to this is it will cause the service be stopped and started but I think that's ok.

Thanks for your questions! I was actually meaning to bring this up with you before implementation... am also up to changing this if we think it's a good idea.

zanderz:
Still, in case someone issues `keybase ctl restart`, won't the watchdog and the restart command each try to spawn a copy?

gabriel:
In cmd_ctl_restart.go, it only forks if fork type isn't watchdog (or launchd)... so I think it's ok. I've also been testing locally (in OS X) using the watchdog instead of launchd to double check this...
https://github.com/keybase/client/blob/master/go/client/cmd_ctl_restart.go#L62

But yeah we'll need to double check all the ctl commands with the new watchdog. Because it's watchdog2 in the client we can test all this without making it live

watchdog/watchdog_test.go
gabriel: Added TestTerminateBeforeWatch

(Also added TestExitOnSuccess which wasn't being tested either.)

watchdog/watchdog_test.go
gabriel: yeah... can use ping.exe with -n timeout which acts same as sleep

gabriel: actually we have a go test binary for stuff like this.. changed to use this one instead

gabriel:
ok.. appveyor wasn't updating this branch, but fixed it and also had to fix some windows tests

I think this might be ready to check-in and for testing

zanderz:
So, the keybase service will be configured to not restart if it ends properly? If not, we'll need it to be I think.
LGTM :thumbsup:

gabriel:
yeah the main service has exit on success turned on.. https://github.com/keybase/client/blob/watchdogv2/go/client/cmd_ctl_watchdog2.go#L51

PR #83, opened 2016-06-02
Use Keybase subdir in appdata (Windows)

gabriel:
@zanderz this matches keybase's (prod) config dir now


zanderz:
Don't want to check for Devel?
LGTM :thumbsup:

gabriel:
Nope updater currently only runs in prod mode... makes everything simpler...

The watchdog too will also only run in prod mode btw

PR #84, opened 2016-06-02
Use temp dir for asset download

gabriel:
This also keeps the file extension which is important for Windows.
Also cleaning up some tmp dirs that weren't being removed after tests.


gabriel:
@zanderz This fixes so the downloaded file doesn't get a weird extension... since on windows it needs to keep the .exe (or .msi) extension

zanderz:
As long as it works, and Windows isn't cranky about .exe/.msi files being placed there - we'll have to try it.
LGTM :thumbsup:

PR #92, opened 2016-06-07
Zanderz/core 2942

zanderz:
I took @chrisnojima 's hta (thanks for that!) and converted it to use plaintext arguments and registry results.
Some of the tests pass on Windows now, too, but not all.
I will check whether the other platforms still build before merging.
One idea I had was to pass the timeout to the .hta so it could close itself instead of being killed.
Anyway, once this is merged I can add it to the new installer. 
@gabriel


keybase/prompter/prompter.hta
```diff
@@ -0,0 +1,108 @@
...
+        window.resizeTo(screen.availWidth/2,screen.availHeight/2);
+        window.moveTo(screen.availWidth/4,screen.availHeight/4);
+     }
+       
+                  
+   </script>  
+<BODY>
+   <div class="relative">
+        <img class="centered_img" src="logo-with-text-dz2-sm.png" />
+
```
chrisnojima: this isn't my hta???

zanderz: Sorry about that, I checked in the wrong one!

gabriel:
I'll take a look

keybase/platform_windows_test.go
```diff
@@ -16,8 +16,16 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type testConfigWindows struct {
+	testConfigPlatform
+}
+
+func (c testConfigWindows) destinationPath() string {
+	return filepath.Join(os.Getenv("GOPATH"), "src/github.com/keybase/go-updater/windows/prompter/prompter.hta")
```
gabriel: Move this into test dir with other resources

keybase/platform_windows.go
```diff
@@ -10,10 +10,22 @@ import (
 	"strings"
 	"time"
 
+	"github.com/kardianos/osext"
 	"github.com/keybase/go-updater"
 	"github.com/keybase/go-updater/command"
 )
 
+// execPath returns the app bundle path where this executable is located
+func (c config) execPath() string {
```
gabriel: i think this is destination path? can move to that?

you could nuke execPath too unless we need it... if so it should be the path to the executable that started the app

keybase/platform_windows_test.go
zanderz: This is the path to the production hta, which will be shipped - there isn't a test hta - where should it live?

keybase/platform_windows.go
zanderz: The intent is for the updater to always find prompter.hta in a subdirectory named "prompter" under the current location from which upd.exe was run. Open to other suggestions of where it lives and how it gets found.

keybase/platform_windows_test.go
gabriel: oh nm, this is real?

gabriel:
We can nuke this in favor of other one?

PR #93, opened 2016-06-07
Windows prompt

gabriel:
See https://github.com/keybase/go-updater/pull/92


test/test.go
```diff
@@ -55,5 +56,9 @@ func output() {
 }
 
 func echo(s string) {
-	fmt.Fprintln(os.Stdout, s)
+	if runtime.GOOS == "windows" {
+		writeToRegistry(s)
```
gabriel: this method won't be found in non-windows

gabriel:
lg2m if build/tests get fixed

gabriel:
We should nuke this PR (in favor of https://github.com/keybase/go-updater/pull/96)?

PR #94, opened 2016-06-08
Remove test channel being hardcoded in update request

gabriel:
We should move to using UIDs (from installIDs)


gabriel:
@keybase/updater-hackers

PR #97, opened 2016-06-09
Fix linux log dir

gabriel:
Was using `Keybase` instead of `keybase` for log, config dir on linux

@oconnor663 

cc @keybase/updater-hackers


oconnor663:
I haven't gotten up to speed on the code here yet. Can we VC and go over it together tomorrow?

oconnor663:
:ship:

PR #99, opened 2016-06-14
Update UI deployment target should be 10.9

gabriel:
@keybase/updater-hackers


cjb:
@gabriel Do we want 10.10 or 10.9?  Are there any actual limitations?

gabriel:
We have a check for >= 10.10 in the Keybase.app itself... (we show error message if they try to run on 10.9)

cjb:
The commits that changes it to 10.9 introduces more fields that weren't being set when you were changing it to 10.10 (DevelopmentTeam, CODE_SIGN_IDENTITY).. what's up with that?

cjb:
:thumbsup:

PR #100, opened 2016-06-15
Some cleanup from windows review

gabriel:
Some general cleanup.

Also testing update prompt with OutPath set in test.

@zanderz 
cc @keybase/updater-hackers


zanderz:
LGTM except CI :thumbsup:

PR #102, opened 2016-06-21
Fixes keybase update source using stale value of Install ID

gabriel:
The `keybase.UpdateSource` (which requests from the keybase.io API endpoint) was getting a copy of config (instead of a config pointer). So any subsequent changes to config (e.g. setting of Install ID) wouldn't be included in requests, and the install_id in the API request was stale. On first installs this means that it would be empty.

However, the install ID does get read correctly after the first update (it sort of fixes itself). This is because when an update occurs, the install ID is saved correctly to config, and at the end of an update, the updater will restart. When the updater starts back up, the `keybase.UpdateSource` reads the new config value and thus can send the correct install ID.

The fix is to have `keybase.UpdateSource` keep a `*config` instead of `config`.

While we did have test coverage to make sure install ID was being saved in the updater tests and in config tests, we didn't have a test to make sure the keybase update source was sending the correct URL params after config changed. This highlights some risks in that we may be lacking "bigger" integration tests that test the whole shebang, in multiple iterations with changing state (e.g. integration tests without anything mocked).

This bug might cause the server to store a lot of install ids that will be unused, that we may need to cleanup. It shouldn't have any other impacts.

This PR also removes a hardcoded "test" channel param which is unused. It bumps the version and includes a ChangeLog.

@keybase/updater-hackers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+
+## ChangeLog
+
+### 0.2.5
+- Fixes bug where keybase update source was using stale value of Install ID
+- Remove hardcoded test channel option (was unused)
--- a/keybase/config.go
+++ b/keybase/config.go
@@ -51,10 +51,10 @@ type store struct {
 }
 
 // newConfig loads a config, which is valid even if it has an error
-func newConfig(appName string, pathToKeybase string, log Log) (config, error) {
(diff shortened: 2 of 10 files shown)


maxtaco:
👍 Thanks for the quick fix!

PR #104, opened 2016-06-23
Process matcher for executable name

gabriel:
- Adds ExecutableName matcher
- Adds more tests around process stuff
- Updates go-ps library
- Increases timeout which could cause flakey test

@jzila Can you take a look?

cc @keybase/updater-hackers


jzila:
LGTM

PR #105, opened 2016-06-27
Unsupported JS on old systems

zanderz:
More scary stuff from testing an unpatched Win7 vm. It would be nice to get this in a build today @gabriel @chrisnojima


gabriel:
👍 

Who are these people running unpatched win7? :)

PR #110, opened 2016-07-26
Request app exit instead of SIGTERM (darwin)

gabriel:
Requests app exit before doing a SIGKILL. I tested both scenarios locally but needs a few days of real testing.

### DIFF
--- a/keybase/platform_darwin.go
+++ b/keybase/platform_darwin.go
@@ -17,7 +17,7 @@ import (
 	"github.com/keybase/go-updater/util"
 )
 
-// destinationPath returns the app bundle path where this executable is located
+// execPath returns the app bundle path where this executable is located
 func (c config) execPath() string {
 	path, err := osext.Executable()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


gabriel:
cc @keybase/updater-hackers

gabriel:
OK this is ready to review

chrisnojima:
seems fine to me, can we get some go 👀  maybe @jzila ?

keybase/platform_darwin.go
```diff
@@ -151,8 +151,16 @@ func (c context) restart(wait time.Duration, delay time.Duration) error {
...
 
-	c.log.Infof("Terminating %s", appProcPath)
-	process.TerminateAll(process.NewMatcher(appProcPath, process.PathContains, c.log), time.Second, c.log)
+	// Request the app to exit (instead of sending a SIGTERM)
+	if appExitResult, err := command.Exec(c.config.keybasePath(), []string{"ctl", "app-exit"}, 5*time.Second, c.log); err != nil {
+		c.log.Warningf("Error requesting app exit: %s (%s)", err, appExitResult.CombinedOutput())
+	}
+	// Give the app time to exit
+	time.Sleep(time.Second)
+	// SIGKILL the app if it failed to exit
```
jzila: why `KILL` and not `TERM`?

jzila:
LGTM except for one question

gabriel:
We do KILL because it we're changing the Keybase.app to do `ctl stop` on SIGTERM, which stops the world (and on updates we want just the app to quit for restart). See https://keybase.atlassian.net/browse/DESKTOP-1552 for more info. There was a long discussion about this topic BTW and this was the consensus.

PR #111, opened 2016-08-02
Command can take env, add KEYBASE_RESTORE_UI to open

gabriel:
This is to support restoring UI state after update,
https://github.com/keybase/client/pull/3695.


gabriel:
@keybase/updater-hackers

chromakode:
:+1:

PR #114, opened 2016-08-24
Prompt user before overriding auto install

zanderz:
The current updater will check whether drivers are being updated and launch the installer non-silently in that case. But if "auto" is on, it skips the dialog and just launches this installer, then kills it after an hour. We need to launch the prompter dialog and kill that instead; killing an installer is particularly bad.
@gabriel @maxtaco

### DIFF
--- a/keybase/config.go
+++ b/keybase/config.go
@@ -36,6 +36,8 @@ type config struct {
 	log Log
 	// store is the config values
 	store store
+	// AutoOverride is whether the current auto setting should be temporarily overridden
+	AutoOverride bool
 }
 
 // store is the config values
@@ -137,6 +139,16 @@ func (c *config) SetUpdateAuto(auto bool) error {
 	c.store.Auto = auto
 	c.store.AutoSet = true
 	return c.save()
+}
+
+// For overriding the current Auto setting
+func (c config) GetUpdateAutoOverride() bool {
+	return c.AutoOverride
+}
+
+func (c *config) SetUpdateAutoOverride(auto bool) error {
+	c.AutoOverride = auto
+	return nil
 }
 
 // GetInstallID is an identifier returned by the API on first update that is a
--- a/keybase/config.go
+++ b/keybase/config.go
@@ -36,8 +36,8 @@ type config struct {
 	log Log
 	// store is the config values
 	store store
-	// AutoOverride is whether the current auto setting should be temporarily overridden
-	AutoOverride bool
+	// autoOverride is whether the current auto setting should be temporarily overridden
+	autoOverride bool
 }
 
 // store is the config values
@@ -143,11 +143,11 @@ func (c *config) SetUpdateAuto(auto bool) error {
 
 // For overriding the current Auto setting
 func (c config) GetUpdateAutoOverride() bool {
-	return c.AutoOverride
+	return c.autoOverride
 }
 
 func (c *config) SetUpdateAutoOverride(auto bool) error {
-	c.AutoOverride = auto
+	c.autoOverride = auto
 	return nil
 }
 
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -80,7 +80,6 @@ type regUninstallGetter func(string, Log) bool
 // among other things generates a log of what it would do. We can parse this
 // for Dokan product code variables, which will be in the registry if the same
 // version is already present.
-//
 func CheckCanBeSilent(path string, log Log, regFunc regUninstallGetter) (bool, error) {
 	tempName := util.TempPath("", "keybaseInstallLayout-")
 
@@ -96,7 +95,7 @@ func CheckCanBeSilent(path string, log Log, regFunc regUninstallGetter) (bool, e
 
 	_, err := command.Exec(path, []string{"/layout", "/quiet", "/log", tempName}, 2*time.Minute, log)
 	if err != nil {
-		log.Errorf("CheckCanBeSilent: Unable to execute %s", path)
+		log.Errorf("CheckCanBeSilent: Unable to execute %s: %s", path, err)
 		return false, err
 	}
 	defer util.RemoveFileAtPath(tempName)
@@ -153,7 +152,7 @@ func checkRebootPending(wow64 bool, log Log) (bool, error) {
 
 	k, err := registry.OpenKey(registry.CURRENT_USER, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce", access)
 	if err != nil {
-		log.Errorf("Error %s opening RunOnce subkeys", err.Error())
+		log.Errorf("Error opening RunOnce subkeys: %s", err)
 		return false, err
 	}
 	defer util.Close(k)
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -61,7 +61,15 @@ func (c config) notifyProgram() string {
 	return ""
 }
 
-func (c context) BeforeUpdatePrompt(update updater.Update, options updater.UpdateOptions) error {
+func (c *context) BeforeUpdatePrompt(update updater.Update, options updater.UpdateOptions) error {
+	// This is the check whether to override an auto update. Unnecessary if
+	// auto update is not on in the first place.
+	_, auto := c.config.GetUpdateAuto()
+	if auto && !c.config.GetUpdateAutoOverride() {
+		if canBeSilent, _ := CheckCanBeSilent(update.Asset.LocalPath, c.log, CheckRegistryUninstallCode); !canBeSilent {
+			c.config.SetUpdateAutoOverride(true)
+		}
+	}
 	return nil
 }
 
@@ -72,8 +80,19 @@ type regUninstallGetter func(string, Log) bool
 // among other things generates a log of what it would do. We can parse this
 // for Dokan product code variables, which will be in the registry if the same
 // version is already present.
+//
 func CheckCanBeSilent(path string, log Log, regFunc regUninstallGetter) (bool, error) {
 	tempName := util.TempPath("", "keybaseInstallLayout-")
+
+	// Also look in the registry whether a reboot is pending from a previous installer.
+	// Not finding the key is not an error.
+	if rebootPending, err := checkRebootPending(false, log); rebootPending && err == nil {
+		return false, nil
+	}
+
+	if rebootPending, err := checkRebootPending(true, log); rebootPending && err == nil {
+		return false, nil
+	}
 
 	_, err := command.Exec(path, []string{"/layout", "/quiet", "/log", tempName}, 2*time.Minute, log)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


keybase/config.go
```diff
@@ -139,6 +139,13 @@ func (c *config) SetUpdateAuto(auto bool) error {
 	return c.save()
 }
 
+// For overriding the current Auto setting
+func (c *config) OverrideAuto() bool {
```
gabriel: I would make a new boolean called `AutoOverride` in config and then check that in the updater (e.g. if Auto || AutoOverride). So that we don't accidentally change/save this.

keybase/platform_windows.go
```diff
@@ -61,7 +61,12 @@ func (c config) notifyProgram() string {
 	return ""
 }
 
-func (c context) BeforeUpdatePrompt(update updater.Update, options updater.UpdateOptions) error {
+func (c *context) BeforeUpdatePrompt(update updater.Update, options updater.UpdateOptions) error {
+	if auto, _ := c.config.GetUpdateAuto(); auto {
```
gabriel: `if c.config.GetUpdateAuto() || c.config.GetUpdateAutoOverride()`

maxtaco:
@gabriel I think you should take this one!

zanderz:
Ping

keybase/config.go
```diff
@@ -48,6 +48,8 @@ type store struct {
 	Auto bool `json:"auto"`
 	// AutoSet is whether a user set the Auto config
 	AutoSet bool `json:"autoSet"`
+	// AutoOverride is whether the current override setting should be temporarily overridden
```
gabriel: "..is whether the current auto setting should be.."

keybase/platform_windows.go
```diff
@@ -191,12 +199,13 @@ func (c context) Apply(update updater.Update, options updater.UpdateOptions, tmp
 		return fmt.Errorf("No asset")
 	}
 	var args []string
-	auto, _ := c.config.GetUpdateAuto()
-	canBeSilent, _ := CheckCanBeSilent(update.Asset.LocalPath, c.log, CheckRegistryUninstallCode)
-	if auto && canBeSilent {
+	_, auto := c.config.GetUpdateAuto()
```
gabriel: Why did you flip the `auto, _`  to `_, auto`?

keybase/platform_windows.go
```diff
@@ -191,12 +199,13 @@ func (c context) Apply(update updater.Update, options updater.UpdateOptions, tmp
...
+	if auto && !c.config.GetUpdateAutoOverride() {
+
 		args = append(args, "/quiet")
 	}
-	_, err := command.Exec(update.Asset.LocalPath, args, time.Hour, c.log)
+	// We really don't want to kill installers, hence 30 days here
+	_, err := command.Exec(update.Asset.LocalPath, args, time.Hour*24*30, c.log)
```
gabriel: Why would the installer take more than an hour?

gabriel:
Some questions

keybase/platform_windows.go
zanderz: The person can leave the reboot prompt up indefinitely.

keybase/platform_windows.go
zanderz: whoops, typo, thanks

keybase/platform_windows.go
zanderz: ...but there is another thing we have to deal with: the person can exit with a reboot pending, in which case the updater also shouldn't re-launch. Still thinking about what to do about that.

gabriel: The timeout is meant to prevent a bad installer from preventing updates so... 

I think maybe this should be max a few days?

What happens again if you kill the installer when it's prompting?

zanderz: One likely scenario is that the reboot prompt gets killed, then the updater launches the install again, which fails repeatedly because a reboot is required.

You are referring to the case where a bad installer is hung? How to distinguish that from an unanswered dialog?

zanderz: actually, I have an idea: we can check the registry whether a reboot is pending - standby for the next commit

zanderz:
Note that the hack I had put in to invoke the new installer turns out to have triggered the reboot. We invoke it with `/layout", "/quiet", "/log"`, which is supposed to only log what it WOULD do, for us to parse out the driver install code. Instead, if a reboot is pending, it will do it immediately with no prompt. I consider this a WIX bug, but this last change will work around it. Note this won't help ppl until the next driver update though.

keybase/config.go
```diff
@@ -48,6 +48,8 @@ type store struct {
 	Auto bool `json:"auto"`
 	// AutoSet is whether a user set the Auto config
 	AutoSet bool `json:"autoSet"`
+	// AutoOverride is whether the current auto setting should be temporarily overridden
+	AutoOverride bool `json:"autoOverride"`
```
gabriel: Should we move this out of the store and into `config` struct?

We never want to store this right?

zanderz:
Looks like I got the wrong branch name too again. Should be maybe CORE-3673 ?

keybase/platform_windows.go
```diff
@@ -125,6 +139,37 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+	if wow64 {
+		access = access | registry.WOW64_32KEY
+	}
+
+	k, err := registry.OpenKey(registry.CURRENT_USER, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce", access)
+	if err != nil {
+		fmt.Printf("Error %s opening RunOnce subkeys\n", err.Error())
+		return
+	}
+	defer k.Close()
```
gabriel: I think you need to use util.Close(k)

This won't pass metalinter

keybase/platform_windows.go
```diff
@@ -125,6 +139,37 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+// Read all the runonce subkeys and find the one with "Keybase" in the name.
+func checkRebootPending(wow64 bool) (result bool) {
+	var access uint32 = registry.ENUMERATE_SUB_KEYS | registry.QUERY_VALUE
+	if wow64 {
+		access = access | registry.WOW64_32KEY
+	}
+
+	k, err := registry.OpenKey(registry.CURRENT_USER, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce", access)
+	if err != nil {
+		fmt.Printf("Error %s opening RunOnce subkeys\n", err.Error())
```
gabriel: Don't std out, use logging..

And log the error at the end... e.g. `log.Errorf("Error opening RunOnce subkeys: %s", err)`.

And don't need `\n`.

keybase/platform_windows.go
```diff
@@ -125,6 +139,37 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
 	return false
 }
 
+// Read all the runonce subkeys and find the one with "Keybase" in the name.
+func checkRebootPending(wow64 bool) (result bool) {
```
gabriel: This should return (bool, error) and then have the caller log error?

I personally stay away from named return values in method signature.

`func checkRebootPending(wow64 bool) (bool, error)`

keybase/platform_windows.go
```diff
@@ -125,6 +139,37 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+
+	for _, param := range params {
+		val, _, err := k.GetStringValue(param)
+
+		if err == nil && strings.Contains(val, "Keybase") {
+			return true
+		}
+	}
+
+	return
```
gabriel: `return false`

keybase/platform_windows.go
```diff
@@ -125,6 +139,37 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+	k, err := registry.OpenKey(registry.CURRENT_USER, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce", access)
+	if err != nil {
+		fmt.Printf("Error %s opening RunOnce subkeys\n", err.Error())
+		return
+	}
+	defer k.Close()
+
+	params, err := k.ReadValueNames(0)
+	if err != nil {
+		fmt.Printf("Can't ReadSubKeyNames %#v", err)
```
gabriel: No printfs

keybase/platform_windows.go
```diff
@@ -125,6 +139,37 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+		return
+	}
+
+	for _, param := range params {
+		val, _, err := k.GetStringValue(param)
+
+		if err == nil && strings.Contains(val, "Keybase") {
```
gabriel: I would change this too:

```
if err != nil {
  return false, err
}
if strings.Contains(val, "Keybase") {
  return true, nil
}
```

The `if err != nil` pattern is very common and so is clearer... Also short circuiting early is better than more complex if statements.

(yeah this is a bit nitpickey.)

gabriel:
You can ignore the test failure if it's `TestTerminateAllPathEqual` on windows. There is a flakeyness.

zanderz:
Yeah I missed that, it belonged in config, not save.

keybase/config.go
```diff
@@ -36,6 +36,8 @@ type config struct {
 	log Log
 	// store is the config values
 	store store
+	// AutoOverride is whether the current auto setting should be temporarily overridden
+	AutoOverride bool
```
gabriel: you can make it private `autoOverride bool`

keybase/platform_windows.go
```diff
@@ -72,9 +80,20 @@ type regUninstallGetter func(string, Log) bool
 // among other things generates a log of what it would do. We can parse this
 // for Dokan product code variables, which will be in the registry if the same
 // version is already present.
+//
```
gabriel: Remove this empty line

keybase/platform_windows.go
```diff
@@ -125,6 +144,41 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+// Read all the runonce subkeys and find the one with "Keybase" in the name.
+func checkRebootPending(wow64 bool, log Log) (bool, error) {
+	var access uint32 = registry.ENUMERATE_SUB_KEYS | registry.QUERY_VALUE
+	if wow64 {
+		access = access | registry.WOW64_32KEY
+	}
+
+	k, err := registry.OpenKey(registry.CURRENT_USER, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce", access)
+	if err != nil {
+		log.Errorf("Error %s opening RunOnce subkeys", err.Error())
```
gabriel: `log.Errorf("Error opening RunOnce subkeys: %s", err)`

keybase/platform_windows.go
```diff
@@ -72,9 +80,20 @@ type regUninstallGetter func(string, Log) bool
...
+		return false, nil
+	}
+
+	if rebootPending, err := checkRebootPending(true, log); rebootPending && err == nil {
+		return false, nil
+	}
+
 	_, err := command.Exec(path, []string{"/layout", "/quiet", "/log", tempName}, 2*time.Minute, log)
 	if err != nil {
 		log.Errorf("CheckCanBeSilent: Unable to execute %s", path)
```
gabriel: `log.Errorf("CheckCanBeSilent: Unable to execute %s: %s", path, err)`

keybase/platform_windows.go
```diff
@@ -125,6 +144,41 @@ func CheckRegistryUninstallCode(productID string, log Log) bool {
...
+func checkRebootPending(wow64 bool, log Log) (bool, error) {
+	var access uint32 = registry.ENUMERATE_SUB_KEYS | registry.QUERY_VALUE
+	if wow64 {
+		access = access | registry.WOW64_32KEY
+	}
+
+	k, err := registry.OpenKey(registry.CURRENT_USER, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce", access)
+	if err != nil {
+		log.Errorf("Error %s opening RunOnce subkeys", err.Error())
+		return false, err
```
gabriel: I guess key not existing isn't really an error but I guess it's ok since we can't distinguish between that and real error

gabriel:
Some minor stuff otherwise 👍 

This is obvious but please test a bunch on windows before we releasing a new updater version. Breaking updates is a huge bummer.

Oh we should bump the updater version number.

PR #119, opened 2016-10-05
Don't steal focus on update prompt (macOS)

gabriel:
This changes the app activation policy to be an NSApplicationActivationPolicyAccessory.
This means the app won't take focus away from other apps when the prompt pops up (and won't have a dock or menu).

It will still be modal, so it will live above other windows.

Also updates project for Xcode 8.


gabriel:
@keybase/react-hackers

MarcoPolo:
👍

PR #120, opened 2016-10-06
Look for Dokan product IDs in updater json

zanderz:
The original technique - running the incoming installer with "/layout", "/quiet", "/log" options - would not work before the new installer was downloaded. Not sure how that one got past me, that's a bad bug. I have tested this by copying a new upd.exe to a previous installation, changing the updater.json on s3 in the test channel to match these changes, and invoking `upd check`. If drivers are the same, update is silent (once again).
@oconnor663 @maxtaco @gabriel

### DIFF
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -4,12 +4,10 @@
 package keybase
 
 import (
-	"bufio"
 	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


keybase/platform_windows.go
```diff
@@ -66,7 +64,16 @@ func (c *context) BeforeUpdatePrompt(update updater.Update, options updater.Upda
 	// auto update is not on in the first place.
 	_, auto := c.config.GetUpdateAuto()
 	if auto && !c.config.GetUpdateAutoOverride() {
-		if canBeSilent, _ := CheckCanBeSilent(update.Asset.LocalPath, c.log, CheckRegistryUninstallCode); !canBeSilent {
+		Dokan86 := ""
```
gabriel: var names are not title cased in go

keybase/platform_windows.go
```diff
@@ -66,7 +64,16 @@ func (c *context) BeforeUpdatePrompt(update updater.Update, options updater.Upda
...
+		Dokan64 := ""
+		for _, prop := range update.Props {
+			if prop.Name == "DokanProductCodeX64" {
```
gabriel: Use switch

keybase/platform_windows.go
```diff
@@ -175,6 +159,26 @@ func checkRebootPending(wow64 bool, log Log) (bool, error) {
...
+		return false, err
+	}
+	defer util.Close(k2)
+	vals, _, err := k2.GetStringsValue("PendingFileRenameOperations")
+	if err != nil {
+		// This is normal if no reboot is pending
+		log.Errorf("Error getting PendingFileRenameOperations: %s", err)
+	} else {
+		for _, val := range vals {
+			if strings.Contains(strings.ToLower(val), "dokan") {
```
gabriel: Is it possible for this to match something not dokan that has the word dokan contained in it? Do we care?

zanderz: The worst that will happen is that the installer will not be launched silently in that case, and the updater prompt will show instead.

PR #122, opened 2016-10-07
Apply autoOverride to prompt also, not just installer launch

zanderz:
This is why we test.
@gabriel @maxtaco

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -229,8 +229,9 @@ func (u *Updater) promptForUpdateAction(ctx Context, update Update, options Upda
 	u.log.Debug("Prompt for update")
 
 	auto, autoSet := u.config.GetUpdateAuto()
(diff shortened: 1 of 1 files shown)


maxtaco:
what's `autoOverride`?

zanderz:
Override the auto update settings, because drivers. IOW, we can't silently update if new drivers are incoming, so show the updater prompt instead of launching an installer once per hour.

zanderz:
PS: I have tested an upgrade scenario with installers built with this

maxtaco:
So autoOveride is specified by the server and disables auto installs?

zanderz:
It is calculated by the client from the new driver product code props in the installer json: if they match (same driver installed as in the update), it can be silent. Before this change, autoOverride only controlled whether the installer was launched, which wasn't what I intended.

maxtaco:
Ok, and it is only one for Windows?

If so, good to merge!

zanderz:
The part that turns it on is windows only, yeah.

PR #124, opened 2016-10-17
Zanderz/core 3864

zanderz:
Matching watchdog changes for drive letter selection and shutting down kbfs properly
@maxtaco @gabriel

### DIFF
--- a/watchdog/watchdog.go
+++ b/watchdog/watchdog.go
@@ -41,10 +41,7 @@ func (p *ProgramNormal) GetExitOn() ExitOn {
 }
 
 func (p *ProgramNormal) DoStop(ospid int, log Log) {
-	matcher := process.NewMatcher(p.Path, process.PathEqual, log)
-	matcher.ExceptPID(ospid)
-	log.Infof("Terminating %s", p.Path)
-	process.TerminateAll(matcher, time.Second, log)
+	StopMatching(p.Path, ospid, log)
 }
 
 // Program is a program at path with arguments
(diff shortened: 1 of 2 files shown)


watchdog/watchdog.go
```diff
@@ -22,12 +22,36 @@ const (
 )
 
 // Program is a program at path with arguments
-type Program struct {
+type ProgramNormal struct {
```
gabriel: Instead of `ProgramNormal` you can use lowercase `program`

gabriel: Also be sure to update the comment... (Linter would catch this)

watchdog/watchdog.go
```diff
@@ -22,12 +22,36 @@ const (
...
 	Path   string
 	Args   []string
 	ExitOn ExitOn
 }
 
+func (p *ProgramNormal) GetPath() string {
```
gabriel: If these are getters, you don't want pointer... e.g. `func (p program) GetPath() string {`

gabriel: The `Get` pattern isn't idiomatic in go...

You should make the args for `program` private and remove `Get` from methods.

For example:

```go
type program struct {
  path string
  ...
}

  func (p program) Path() string {
    return p.path
  }
```

watchdog/watchdog.go
```diff
@@ -22,12 +22,36 @@ const (
...
+func (p *ProgramNormal) DoStop(ospid int, log Log) {
+	StopMatching(p.Path, ospid, log)
+}
+
+// Program is a program at path with arguments
+type Program interface {
+	GetPath() string
+	GetArgs() []string
+	GetExitOn() ExitOn
+	DoStop(ospid int, log Log)
```
gabriel: I would call this:

`StopMatching(exceptPid int, log Log)`

watchdog/watchdog.go
```diff
@@ -46,15 +70,19 @@ func Watch(programs []Program, restartDelay time.Duration, log Log) error {
 	return nil
 }
 
+func StopMatching(path string, ospid int, log Log) {
```
gabriel: you can make this a private method

`func stopMatching(path string, exceptPid int, log Log) `

watchdog/watchdog.go
```diff
@@ -46,15 +70,19 @@ func Watch(programs []Program, restartDelay time.Duration, log Log) error {
...
+	matcher := process.NewMatcher(path, process.PathEqual, log)
+	matcher.ExceptPID(ospid)
```
gabriel: ospid should be `exceptPid` cause otherwise very confusing

PR #125, opened 2016-11-11
Try privileged keybase uninstall on rename error

gabriel:
This allows the updater to replace the Keybase.app, even if the Keybase.app is installed by another user.

If we get an `os.LinkError, Op:"rename"` for the Keybase.app when trying to apply, we'll try to do a `keybase uninstall --components=app` which can remove `/Applications/Keybase.app` via the privileged helper.

### DIFF
--- a/keybase/platform_darwin.go
+++ b/keybase/platform_darwin.go
@@ -231,6 +231,12 @@ func (c context) check(sourcePath string, destinationPath string) error {
 	return nil
 }
 
+func (c context) apply(localPath string, destinationPath string, tmpDir string) error {
+	// The file name we unzip over should match the (base) file in the destination path
+	filename := filepath.Base(destinationPath)
+	return util.UnzipOver(localPath, filename, destinationPath, c.check, tmpDir, c.log)
+}
+
 func (c context) Apply(update updater.Update, options updater.UpdateOptions, tmpDir string) error {
 	if update.Asset == nil {
 		return fmt.Errorf("No asset")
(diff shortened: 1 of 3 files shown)


service/main.go
```diff
@@ -51,12 +52,11 @@ func defaultAppName() string {
 }
 
 func run(f flags) {
-	ulog := logger{}
-
 	if f.version {
-		ulog.Infof("%s\n", updater.Version)
+		fmt.Printf("%s\n", updater.Version)
```
gabriel: This is unrelated but wanted to fix how we output the version. This allows 
```sh
updater_version=`updater -version`
```
to work in scripts

gabriel:
@chrisnojima for conceptual review?
@jzila for go code review?

keybase/platform_darwin.go
```diff
@@ -231,17 +231,37 @@ func (c context) check(sourcePath string, destinationPath string) error {
...
 		return fmt.Errorf("No asset")
 	}
 	localPath := update.Asset.LocalPath
 	destinationPath := options.DestinationPath
 
-	// The file name we unzip over should match the (base) file in the destination path
-	filename := filepath.Base(destinationPath)
-	if err := util.UnzipOver(localPath, filename, destinationPath, c.check, tmpDir, c.log); err != nil {
-		return err
+	if err := c.apply(localPath, destinationPath, tmpDir); err != nil {
```
jzila: There's a lot of activity in this block that requires reading the whole block to understand any individual code path. If we make the return scenarios explicit, we can avoid that.
```go
err := c.apply(localPath, destinationPath, tmpDir)
switch err.(type) {
case nil:
case *os.LinkError:
    if err.Op == "rename" && err.Old == "/Applications/Keybase.app" {
        c.log.Infof("The error was a problem renaming (moving) the app, let's trying uninstalling the app via keybase uninstall which has more privileges")
        _, uninstallErr := command.Exec(c.config.keybasePath(), []string{"uninstall", "--components=app"}, 10*time.Second, c.log)
        if uninstallErr != nil {
            c.log.Errorf("Error trying to uninstall the app: %s", uninstallErr)
            return err
        } else {
            c.log.Infof("We uninstalled the app, let's try to apply again")
            err := c.apply(localPath, destinationPath, tmpDir)
            if err != nil {
                return err
            }
        }
    } else {
        return err
    }
default:
    return err
}
```

gabriel:
Yes yes totally... way better. Be explicit about returns and don't re-use/set error var.

PR #126, opened 2016-12-02
use SHGetKnownFolderPath, local instead of roaming

zanderz:
Use system APIs instead of relying on the environment variable. Matching PRs in client and kbfs.
@maxtaco @gabriel


keybase/platform_windows.go
```diff
@@ -30,9 +81,9 @@ func (c config) destinationPath() string {
 
 // Dir returns where to store config and log files
 func Dir(appName string) (string, error) {
-	dir := os.Getenv("APPDATA")
-	if dir == "" {
-		return "", fmt.Errorf("No APPDATA env set")
+	dir, err := LocalDataDir()
+	if err != nil || dir == "" {
```
gabriel: I think you want to break this up...

```
if err != nil { return "", err }
if dir == "" { return "", fmt.Errorf("No APPDATA env set") }
```

keybase/platform_windows.go
```diff
@@ -5,19 +5,70 @@ package keybase
...
+type GUID struct {
+	Data1 uint32
+	Data2 uint16
+	Data3 uint16
+	Data4 [8]byte
+}
+
+// F1B32785-6FBA-4FCF-9D55-7B8E7F157091
+var (
+	FOLDERID_LocalAppData = GUID{0xF1B32785, 0x6FBA, 0x4FCF, [8]byte{0x9D, 0x55, 0x7B, 0x8E, 0x7F, 0x15, 0x70, 0x91}}
```
gabriel: maybe this varname could be FolderIDLocalAppData

keybase/platform_windows.go
```diff
@@ -5,19 +5,70 @@ package keybase
...
 
 	"github.com/kardianos/osext"
 	"github.com/keybase/go-updater"
 	"github.com/keybase/go-updater/command"
 	"github.com/keybase/go-updater/util"
+	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/registry"
 )
 
+type GUID struct {
```
gabriel: I think a lot of these structs methods could be private? ... e.g. `type guid struct`

keybase/platform_windows.go
zanderz: This name is directly from the API, it seems better that it matches: https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457(v=vs.85).aspx

keybase/platform_windows.go
```diff
@@ -5,19 +5,71 @@ package keybase
...
+var (
+	modShell32               = windows.NewLazySystemDLL("Shell32.dll")
+	modOle32                 = windows.NewLazySystemDLL("Ole32.dll")
+	procSHGetKnownFolderPath = modShell32.NewProc("SHGetKnownFolderPath")
+	procCoTaskMemFree        = modOle32.NewProc("CoTaskMemFree")
+)
+
+func coTaskMemFree(pv uintptr) {
+	syscall.Syscall(procCoTaskMemFree.Addr(), 1, uintptr(pv), 0, 0)
+	return
```
gabriel: Don't need this return?

keybase/platform_windows.go
```diff
@@ -5,19 +5,71 @@ package keybase
...
+		return "", errors.New("can't get FOLDERID_RoamingAppData")
+	}
+
+	defer coTaskMemFree(pszPath)
+
+	// go vet: "possible misuse of unsafe.Pointer"
+	folder := syscall.UTF16ToString((*[1 << 16]uint16)(unsafe.Pointer(pszPath))[:])
+
+	if len(folder) == 0 {
+		return "", errors.New("can't get AppData directory")
```
gabriel: I think most of our errors are title cased... "Can't...

keybase/platform_windows.go
```diff
@@ -5,19 +5,71 @@ package keybase
...
+	syscall.Syscall(procCoTaskMemFree.Addr(), 1, uintptr(pv), 0, 0)
+	return
+}
+
+func getDataDir(id guid) (string, error) {
+
+	var pszPath uintptr
+	r0, _, _ := procSHGetKnownFolderPath.Call(uintptr(unsafe.Pointer(&id)), uintptr(0), uintptr(0), uintptr(unsafe.Pointer(&pszPath)))
+	if r0 != 0 {
+		return "", errors.New("can't get FOLDERID_RoamingAppData")
```
gabriel: I think most of our errors are title cased... "Can't...

gabriel:
lg2m with some comments...

Bump the version too?

PR #127, opened 2016-12-22
Increase timeout when getting Keybase version

zanderz:
and use 10 second timeout instead of 5 - getting the version has failed on Windows before.
@maxtaco @gabriel

### DIFF
--- a/keybase/config.go
+++ b/keybase/config.go
@@ -182,7 +182,7 @@ func (c config) keybasePath() string {
 }
 
 func (c config) keybaseVersion() string {
-	result, err := command.Exec(c.keybasePath(), []string{"version", "-S"}, 5*time.Second, c.log)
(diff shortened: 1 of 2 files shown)


gabriel:
I think this is ok..

So you're saying a user had a current version of keybase but it was timing out or erroring on the version command? Was the computer being slow or was it something else, just curious.

We could also bump up the timeout to 30s instead of retrying if it was only a timeout issue.

gabriel:
Feel free to bump version

zanderz:
Hard to say, here are logs: https://github.com/keybase/keybase-issues/issues/2695
I also have a speculative service PR to match: https://github.com/keybase/client/pull/5202

gabriel:
Ah ok, if it's a keybase causing computer to run slow... can we bump up the timeout to like 20 or 30 seconds and not do retries?

Also I posted a comment in the other PR.... It appears that any `keybase` CLI command is triggering identifies to occur and not necessarily on an hourly basis.

PR #129, opened 2017-02-22
Stop, apply, start (darwin)

gabriel:


keybase/platform_test.go
```diff
@@ -21,12 +21,6 @@ func (c testConfigPlatform) promptProgram() (command.Program, error) {
 	if programPath == "" {
 		programPath = filepath.Join(os.Getenv("GOPATH"), "bin", "test")
 	}
-	if len(args) == 0 {
```
gabriel: Moved this default arg into the 1 test that was using it (should be with the test, not a default).

command/command.go
```diff
@@ -47,7 +48,14 @@ type Result struct {
 
 // CombinedOutput returns Stdout and Stderr as a single string.
 func (r Result) CombinedOutput() string {
-	return fmt.Sprintf("[Stdout] %s\n[Stdrr] %s", r.Stdout.String(), r.Stderr.String())
+	strs := []string{}
```
gabriel: Makes the output more readable in logs.

process/process_test.go
```diff
@@ -168,9 +168,9 @@ func testTerminateAll(t *testing.T, path string, matcher Matcher, numProcs int)
 		defer cleanupProc(cmd, "")
 	}
 
-	time.Sleep(1 * time.Second)
+	time.Sleep(time.Second)
 
-	terminatePids := TerminateAll(matcher, 5*time.Second, testLog)
+	terminatePids := TerminateAll(matcher, time.Second, testLog)
```
gabriel: This was slowing down tests

gabriel:
@jzila You've reviewed go updater in the past, do you want to review this?

gabriel:
If auto update is on and KBFS is not in use, it will kill it same way as before. We need to add a "Can we update you?" confirm dialog now that chat is live, otherwise yeah, it could re-start you.

I'll make an issue for that.

PR #134, opened 2017-07-13
Faster check for update on startup

gabriel:
Currently the tickDuration is 1h between checks. The ticker doesn't trigger a check until after duration. This makes a check before starting ticker.

The reasoning for not doing this initially was to be really conservative, in case of a crash during check, we don't spam the API server. 

On startup, we use a touched file mod time to determine if we should attempt a check right away.


cjb:
CI builds are failing.

Do we want to do this?  It would leave us vulnerable to a DoS from clients whose updaters are crashing immediately and then rechecking immediately when they're respawned.

Maybe it should be more likely wait 5 minutes before first check, and then an hour between subsequent checks?

cjb:
@jzila unless the leveldb is failing to be updated because we're crashing at startup or misconfigured :)

jzila:
@cjb right, so we'd trigger the check _after_ the levelDB is updated.

keybase/config.go
```diff
@@ -107,6 +108,44 @@ func (c config) path() (string, error) {
...
+}
+
+// SetLastUpdateTime touches file to set last update time.
+func (c config) SetLastUpdateTime() {
+	configDir, err := Dir(c.appName)
+	if err != nil {
+		c.log.Errorf("Error getting config dir: %s", err)
+		return
+	}
+	path := filepath.Join(configDir, "updater.last")
```
jzila: duplicated code with the above function through this line.

PR #135, opened 2017-07-31
Prompt for apply

gabriel:
Always prompt to apply. If in use, then use the previous prompt.

Otherwise show an apply prompt:
2017-07-31 at 2 31 pm


cjb:
(Please don't merge this until we've agreed on whether it's the UX we want.)

gabriel:
Yeah just a proof of concept

PR #138, opened 2017-10-30
Add a pid file when running as a service

patrickxb:
This creates an updater.pid file in the CacheDir.


patrickxb:
I think I fixed all the cross-platform issues...

patrickxb:
(my launchd fallback depends on this change, so I'd like to get this in first)

maxtaco:
Should anything else go into the CacheDir?

maxtaco:
(ignore my comment)

PR #140, opened 2017-12-06
real Windows arch

zanderz:
Turns out we were just sending the arch used for building the binary, which doesn't seem useful. @jzila suggests we may want it still, and to use a new field for the actual user's arch. I think we could worry about that later. @maxtaco ?
@keybase/updater-hackers

### DIFF
--- a/keybase/platform_darwin.go
+++ b/keybase/platform_darwin.go
@@ -4,8 +4,10 @@
 package keybase
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
 	"os/user"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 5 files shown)


keybase/platform_darwin.go
```diff
@@ -95,6 +96,11 @@ func (c config) osVersion() string {
 	return strings.TrimSpace(result.Stdout.String())
 }
 
+// TODO: this only tells us the compile time arch
+func (c config) osArch() string {
+	return runtime.GOARCH
```
jzila: This can be:
```go
cmd := exec.Command("uname", "-m")
var buf bytes.Buffer
cmd.Stdout = &buf
err := cmd.Run()
if err != nil {
    return runtime.GOARCH
}
return buf.String()
```

keybase/platform_linux.go
```diff
@@ -66,6 +67,11 @@ func (c config) osVersion() string {
 	return strings.TrimSpace(result.Stdout.String())
 }
 
+// TODO: this only tells us the compile time arch
+func (c config) osArch() string {
+	return runtime.GOARCH
```
jzila: Same comment as Darwin.

PR #142, opened 2018-01-07
forgot to actually send Arch to update server

zanderz:
Does anything else need to happen on the server side?
@maxtaco @jzila


maxtaco:
Ask @cjb and @oconnor663

oconnor663:
(Linux doesn't go anywhere near this code, so I'm out of the loop.)

maxtaco:
@oconnor663  -- I meant the server-side code in web/pkg.iced, weren't you there recently?

zanderz:
It seems like the server accepts it as an optional parameter already: https://github.com/keybase/keybase/blob/master/web/pkg.iced#L59

oconnor663:
Not since 2016 :p

zanderz:
so, guys, where does that leave us @oconnor663 @maxtaco ?

PR #146, opened 2018-02-22
osx: stop the redirector before an upgrade

strib:
Note that this shouldn't be merged until keybase/client#10648 is merged into client master.  But it's good for review now.

Issue: KBFS-2787

### DIFF
--- a/keybase/platform_darwin.go
+++ b/keybase/platform_darwin.go
@@ -184,7 +184,7 @@ func (c context) stop() error {
 	}
 
 	// Stop the redirector so it can be upgraded for all users.
-	_, redirectorErr := command.Exec(c.config.keybasePath(), []string{"uninstall", "--components=redirector", fmt.Sprintf("--source-path=%s", sourcePath)}, time.Minute, c.log)
+	_, redirectorErr := command.Exec(c.config.keybasePath(), []string{"uninstall", "--components=redirector"}, time.Minute, c.log)
 	if redirectorErr != nil {
 		c.log.Warningf("Error stopping the redirector: %s", redirectorErr)
 	}
--- a/keybase/platform_darwin.go
+++ b/keybase/platform_darwin.go
@@ -183,6 +183,12 @@ func (c context) stop() error {
 		c.log.Warningf("Error requesting app exit: %s", appExitErr)
 	}
 
+	// Stop the redirector so it can be upgraded for all users.
+	_, redirectorErr := command.Exec(c.config.keybasePath(), []string{"uninstall", "--components=redirector", fmt.Sprintf("--source-path=%s", sourcePath)}, time.Minute, c.log)
+	if redirectorErr != nil {
+		c.log.Warningf("Error stopping the redirector: %s", redirectorErr)
+	}
+
 	// Stop services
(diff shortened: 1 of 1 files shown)


keybase/platform_darwin.go
```diff
@@ -183,6 +183,12 @@ func (c context) stop() error {
 		c.log.Warningf("Error requesting app exit: %s", appExitErr)
 	}
 
+	// Stop the redirector so it can be upgraded for all users.
```
jzila: It looks like you're uninstalling the redirector first, not just stopping it. That's fine, but the comments and error messages should reflect what's actually happening.

Also, what happens if the upgrade fails?

jzila: One more question: will this work before a redirector exists?

strib: > It looks like you're uninstalling the redirector first, not just stopping it. That's fine, but the comments and error messages should reflect what's actually happening.

Under the covers, `--uninstall-redirector` just calls `stopRedirector` in the root helper.  It's following the naming convention of everything else, including `--uninstall-kbfs` which just stops KBFS but doesn't actually remove anything from the system.

> Also, what happens if the upgrade fails?

Then the redirector will remain shut down until the next time the app starts up.  I think this matches the behavior of shutting down the rest of the application before the upgrade happens.

> One more question: will this work before a redirector exists?

Nope. That's why I mention in the PR description that I won't merge this until the redirector changes are merged into master.

keybase/platform_darwin.go
```diff
@@ -183,6 +183,12 @@ func (c context) stop() error {
...
+	_, redirectorErr := command.Exec(c.config.keybasePath(), []string{"uninstall", "--components=redirector"}, time.Minute, c.log)
```
songgao: Is `c.config.keybasePath()` the `keybase` binary? Out of curiosity I did `keybase uninstall` and it seemed to be removing a bunch of binaries from my system as part of uninstalling the app. So it sounds like `keybase uninstall` is different from "uninstalling" a mount?

strib: > Is c.config.keybasePath() the keybase binary?

Yep.

> So it sounds like keybase uninstall is different from "uninstalling" a mount?

`keybase uninstall` without any flags attempts to do a full uninstall.  But if you give it `--components`, you can do smaller things, like stopping KBFS, or uninstalling just the fuse module, or deleting just the mount directory.  So I added a component for stopping the redirector.

songgao: Got you. Thanks for explaining it. So there are two levels of "uninstall":

1. Stop the redirector and removing the dir that corresponds to the mount point.

2. Remove the keybase-redirector itself.

It seems uninstalling the `redirector` component is 1. I suppose we still need 2 somewhere?

strib: > It seems uninstalling the redirector component is 1. I suppose we still need 2 somewhere?

Maybe, though it would be as part of one of the other PRs in client.  I can test it out and see if that binary is treated differently from the others.

songgao: Got it; thanks!

strib:
Shit, merged the wrong thing.  Reverting.

PR #150, opened 2018-03-08
updater can't handle .msi after all

zanderz:
what a bummer. At least this is tested for if/when we want to use an .msi, but obvs we need to wait until updaters are all updated.

### DIFF
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -208,13 +208,17 @@ func (c context) Apply(update updater.Update, options updater.UpdateOptions, tmp
 	if update.Asset == nil || update.Asset.LocalPath == "" {
 		return fmt.Errorf("No asset")
 	}
-
+	runCommand := update.Asset.LocalPath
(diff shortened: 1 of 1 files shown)


maxtaco:
What is the roll-out strategy going to be? If I turn my machine off before I get this update, and then I miss a whole bunch of updates, and then I get an `.msi` on the only updater, won't my install be bricked?

It seems the updater has to tell the server it wants an .msi or an .exe, and we need to support both on the server for a while.  

Are there other possibilities?

zanderz:
Sure, one very likely possibility is that we never ship an .msi at all, like if we switch away from wix completely.
Another could be that we leave the last exe upgrade in place for such late users, and starting with msi the updater looks elsewhere, or in another channel, and the last .exe version will know how to upgrade to .msi.

maxtaco:
I reviewed this one before I reviewed the other PR. Seems like a good strategy to put support in now that we wish we had...

PR #151, opened 2018-03-20
listen for inactive gui for 3 timeouts

zanderz:
first attempt. Notice it only works in silent update mode.
Haven't tried it on Mac but it should be harmless, I'd like to merge.

### DIFF
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -30,10 +30,13 @@ type guid struct {
 	Data4 [8]byte
 }
 
-// FOLDERID_LocalAppData
-// F1B32785-6FBA-4FCF-9D55-7B8E7F157091
 var (
+	// FOLDERID_LocalAppData
+	// F1B32785-6FBA-4FCF-9D55-7B8E7F157091
 	folderIDLocalAppData = guid{0xF1B32785, 0x6FBA, 0x4FCF, [8]byte{0x9D, 0x55, 0x7B, 0x8E, 0x7F, 0x15, 0x70, 0x91}}
+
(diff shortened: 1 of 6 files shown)


keybase/platform_windows.go
```diff
@@ -34,6 +34,7 @@ type guid struct {
 // F1B32785-6FBA-4FCF-9D55-7B8E7F157091
```
jzila: Please move this comment to before the `folderIDLocalAppData` constant, and make an equivalent for `folderIDRoamingAppData`.

updater.go
```diff
@@ -256,6 +267,36 @@ func (u *Updater) promptForUpdateAction(ctx Context, update Update, options Upda
...
+		return false, nil
+	}
+
+	rawState, err := util.ReadFile(ctx.GetAppStatePath())
+	if err != nil {
+		u.log.Warningf("Error reading GUI state - proceeding", err)
+		return false, nil
+	}
+	
+    guistate := guiAppState{}
```
jzila: whitespace

updater.go
```diff
@@ -256,6 +267,36 @@ func (u *Updater) promptForUpdateAction(ctx Context, update Update, options Upda
...
+}
+
+func (u *Updater) checkUserActive(ctx Context) (bool, error) {
+
+	if u.guiBusyCount >= 3 {
+		u.log.Warningf("Waited for GUI %d times - ignoring busy", u.guiBusyCount)
+		return false, nil
+	}
+
+	rawState, err := util.ReadFile(ctx.GetAppStatePath())
```
jzila: Where is this file written?

PR #155, opened 2018-03-29
Revert "listen for inactive gui for 3 timeouts"

cjb:
@keybase/react-hackers CC @chrisnojima 

Reverts keybase/go-updater#151

This breaks `keybase update check` from working on macOS if the GUI's been active recently.  But more importantly, I don't think we intended to ship this feature yet, because the rest of the updater work isn't done.


zanderz:
Why not just skip the active detection in the "check" case?

PR #157, opened 2018-04-04
Download new builds before prompting for their installation

cjb:
@keybase/updater-hackers 

Remove most of the delay between the update prompt and the app restart, by performing the long download before prompting.


updater_test.go
```diff
@@ -269,13 +269,11 @@ func TestUpdaterDownloadError(t *testing.T) {
 
 	require.NotNil(t, ctx.errReported)
 	assert.Equal(t, ctx.errReported.(Error).errorType, DownloadError)
-	assert.Equal(t, "deadbeef", ctx.updateReported.InstallID)
```
cjb: You're only going to have `updateReported` if the user's chosen an action at the prompt, and here we get a download error, which happens before the prompt, so just remove this part of the test.

updater_test.go
```diff
@@ -269,13 +269,11 @@ func TestUpdaterDownloadError(t *testing.T) {
...
-	assert.Equal(t, "cafedead", ctx.updateReported.RequestID)
 	assert.False(t, ctx.successReported)
 }
 
 func TestUpdaterCancel(t *testing.T) {
-	testServer := testServerForError(t, fmt.Errorf("cancel"))
```
cjb: If we want to test the prompt action flow, we now need to go through with getting a download first.

updater_test.go
```diff
@@ -289,7 +287,7 @@ func TestUpdaterCancel(t *testing.T) {
 }
 
 func TestUpdaterSnooze(t *testing.T) {
-	testServer := testServerForError(t, fmt.Errorf("snooze"))
```
cjb: If we want to test the prompt action flow, we now need to go through with getting a download first.

joshblum:
(approved via @cjb in chat)

PR #161, opened 2018-04-19
allow auto update with check command

zanderz:


mlsteele:
What is `autoOverride`? I see that it comes from `config struct`. Is it only for tests?

updater.go
```diff
@@ -129,6 +129,8 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
 		return update, promptErr(fmt.Errorf("Unknown prompt error"))
 	case UpdateActionContinue:
 		// Continue
+	case UpdateActionUIBusy:
+		// Continue (no need to report this to the server)
```
mlsteele: I think the point of finding out `UIBusy` is to _not_ do the update yet. So this should be more like `case UpdateActionError`, right?

updater.go
```diff
@@ -232,15 +234,18 @@ func (u *Updater) promptForUpdateAction(ctx Context, update Update, options Upda
...
+	if auto && !autoOverride {
+		if !ctx.IsCheckCommand() {
+			isActive, err := u.checkUserActive(ctx)
+			if isActive {
+				err = fmt.Errorf("GUI is active, try later")
+				return UpdateActionUIBusy, err
+			}
+			if err != nil {
+				return UpdateActionError, err
+			}
```
mlsteele: Woah do you think there's a reason this `if isActive` and `if err != nil` was in the wrong order before?

mlsteele: Now that you've changed `checkUserActive` to actually return errors instead of just returning `err=nil` all the time. If that thing _fails_ the proper response is probably to go ahead and update. A faulty gui writing bogus app-state files shouldn't block upgrades forever.

What about something like this where we just ignore errors from `checkUserActive`?
```
isActive, err := u.checkUserActive(ctx)
if err == nil && isActive {
    return UpdateActionUIBusy, nil
}
u.guiBusyCount == 0
```

error.go
```diff
@@ -15,6 +15,8 @@ const (
 	CancelError ErrorType = "cancel"
 	// ConfigError is for errors reading/saving config
 	ConfigError ErrorType = "config"
+	// ConfigError is for when the GUI is active
+	GUIBusyError ErrorType = "guiBusy"
```
mlsteele: Now this dude's unused

updater.go
zanderz: ah of course, sorry. I added to the test for this case.

error.go
zanderz: It's used in IsGuiBusy() to test for this case in report()

mlsteele: oh yeah my bad

PR #163, opened 2018-06-04
bump version

zanderz:
Seems to have stayed 2.13 since October, though .msi capability on Windows was added in March, etc.


maxtaco:
i'd go to 0.3.0

PR #164, opened 2018-06-22
Do some detection of messed up installs and some remediation

zanderz:
Earlier experimental versions actually ran uninstall, but that is too dangerous to do automatically IMHO. Let's start by shipping this.
@keybase/react-hackers

### DIFF
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -386,7 +386,7 @@ func (c context) Apply(update updater.Update, options updater.UpdateOptions, tmp
 }
 
 // Note that when a Windows installer runs, it kills the running updater, even
-// before AfterApply() ruhs
+// before AfterApply() runs
 func (c context) AfterApply(update updater.Update) error {
 	return nil
 }
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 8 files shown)


keybase/platform_windows.go
```diff
@@ -211,10 +212,149 @@ func (c context) PausedPrompt() bool {
...
+
+	UIDs, err := k.ReadSubKeyNames(-1)
+	if err != nil {
+		c.log.Infof("Error reading subkeys: %s\n", err.Error())
+		return
+	}
+	for _, UID := range UIDs {
+		componentsKey, err := registry.OpenKey(k, UID+"\\Components", readAccess)
+		if err != nil {
+			c.log.Infof("Error opening subkey %s: %s\n", UID+"\\Components", err.Error())
```
maxtaco: This error is just dropped and we charge right through. Is that OK?

keybase/platform_windows.go
```diff
@@ -211,10 +212,149 @@ func (c context) PausedPrompt() bool {
...
+			if err != nil {
+				if componentKeyAccess == writeAccess {
+					c.log.Infof("Error opening subkey for writing %s: %s\n", componentKeyName, err.Error())
+					componentKeyAccess = readAccess
+					componentKey, err = registry.OpenKey(componentsKey, componentKeyName, componentKeyAccess)
+				}
+				// TODO: No need to list all the components we couldn't open in write mode.
+				// This is expected when run without elevated permissions.
+				if err != nil {
+					continue
```
maxtaco: does it make sense to warn in the case of this error?

keybase/platform_windows.go
```diff
@@ -211,10 +212,149 @@ func (c context) PausedPrompt() bool {
...
+						}
+					}
+					foundKeybase = true
+				}
+			}
+			componentKey.Close()
+		}
+		componentsKey.Close()
+	}
+	return
```
maxtaco: Change to `return result`

keybase/platform_windows.go
```diff
@@ -211,10 +212,149 @@ func (c context) PausedPrompt() bool {
...
+			c.log.Infof("Removing %s", f)
+			if err = os.Remove(f); err != nil {
+				c.log.Infof("Error removing file: %s", err.Error())
+			}
+		}
+	}
+}
+
+// DeepClean is called when a faulty upgrade has been detected
+// Eventually we may need to preform full uninstalls but that is kind of riskyk
```
maxtaco: `riskyk` misspelling

keybase/platform_windows.go
```diff
@@ -211,10 +212,149 @@ func (c context) PausedPrompt() bool {
 	return false
 }
 
+// checkRegistryComponents will fall back to read-only if it can't do deletion,
+// returning true if either entries were deleted, or multiple Keybase products were found
+func (c context) checkRegistryComponents(wow64, readonly bool) (result bool) {
```
maxtaco: `readonly` is only ever set to `true`.  Can you get rid of it?

maxtaco: It's bugprone to have functions take a long list of boolean arguments.

keybase/platform_windows.go
zanderz: Don't think so!

maxtaco:
@zanderz just happened across this tab. LMK when it's ready for another look

zanderz:
yes I think it's good for another look now, thanks

PR #165, opened 2018-07-05
7 day snooze on Windows

zanderz:
Now we send snooze_duration too.
@keybase/react-hackers 

snooze7


maxtaco:
the text is fine, but this isn't fully plumbed through on kbweb, so i'm guessing it's not working?

maxtaco:
(needs a two-line fix to supply `snooze_duration` instead of `0` in `_write_db`

PR #168, opened 2018-07-24
Fix a bug that kept updates from happening as long as GUI was busy

zanderz:
possibly introduced when the command line check logic was added?

@keybase/react-hackers


zanderz:
this is a 1 line change

PR #171, opened 2018-08-30
add need-update cmd to updater

songgao:
This adds a `need-update` subcommand to the updater, so we can call the command to figure out if we should show the "out of date" banner to user.


songgao:
Hmm do we need to build a KeybaseUpdater to include this?

PR #172, opened 2018-08-31
Revert "Revert "Fix a bug that kept updates from happening as long as…

zanderz:
… GUI was busy" (#169)"

This reverts commit 1abc038d7dd519b6ca0f0e6d4dda2f36e7bf4481.

@keybase/react-hackers

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -15,7 +15,7 @@ import (
 )
 
 // Version is the updater version
-const Version = "0.3.2"
+const Version = "0.3.1"
 
 // Updater knows how to find and apply updates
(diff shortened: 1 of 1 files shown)


updater.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // Version is the updater version
-const Version = "0.3.2"
+const Version = "0.3.1"
```
chrisnojima: this needs to go up not down right

zanderz: ha, whoops!

PR #173, opened 2018-09-05
wasn't sending up snooze duration after all

zanderz:
@keybase/react-hackers

### DIFF
--- a/keybase/report.go
+++ b/keybase/report.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/keybase/go-updater"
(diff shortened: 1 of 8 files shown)


zanderz:
@maxtaco we should either get this in or disable 7 day snooze before shipping Monday

maxtaco:
Ah man, I really don't know this code that well.  @cjb do you know this code at all?

keybase/report.go
```diff
@@ -37,17 +38,20 @@ func (c context) reportError(err error, update *updater.Update, options updater.
...
 
-func (c context) reportAction(action updater.UpdateAction, update *updater.Update, options updater.UpdateOptions, uri string, timeout time.Duration) error {
+func (c context) reportAction(actionResponse updater.UpdatePromptResponse, update *updater.Update, options updater.UpdateOptions, uri string, timeout time.Duration) error {
 	data := url.Values{}
-	data.Add("action", action.String())
+	data.Add("action", actionResponse.Action.String())
 	autoUpdate, _ := c.config.GetUpdateAuto()
 	data.Add("auto_update", util.URLValueForBool(autoUpdate))
+	if actionResponse.SnoozeDuration > 0 {
+		data.Add("snooze_duration", strconv.Itoa(actionResponse.SnoozeDuration))
```
maxtaco: `fmt.Sprintf("%d",...)` is what I typically use

zanderz:
The snooze duration wasn't being passed around with the basic response. Maybe we should have just had a second flavor of snooze instead of an extra value for the duration, but it seemed to make sense to pass the whole response object so we could get that duration if we want.

PR #176, opened 2018-10-02
slight bug in cleanup mode

zanderz:
@maxtaco there's a bug in https://github.com/keybase/go-updater/pull/164 : in aggressive cleanup mode it deletes all the executables, including the runquiet launcher, so `keybase ctl stop` fails, so the upgrade can't overwrite the files it needs to. In non-silent mode, you get a dialog saying it can't continue because certain programs are still running. Silent mode probably fails.

Fix is to try closing keybase before the forced cleanup, and not do silent mode in this case. I tested this.

Not sure who else to put on this review.

### DIFF
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -305,6 +305,13 @@ func (c context) deleteProductFiles() {
 		c.log.Infof("Error getting Keybase directory: %s", err.Error())
 		return
 	}
+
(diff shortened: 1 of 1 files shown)


mmaxim:
This is the exact bug I see on my Windows install.

zanderz:
I was thinking about this some more - maybe in silent mode the updater should explicitly issue a `keybase ctl stop` to make things smoother?
@mmaxim do you have a recent log send from that condition? Thanks.

PR #177, opened 2018-10-09
deleting executables on failed upgrade was too aggressive

zanderz:
Deleting executables was a desperate hack and it is causing more problems. Next thing to try could be to always do `keybase ctl stop` on update instead of depending on the .msi to do so.

### DIFF
--- a/keybase/platform_windows.go
+++ b/keybase/platform_windows.go
@@ -299,7 +299,7 @@ func (c *ComponentsChecker) checkRegistryComponents() (result bool) {
 	return result
 }
 
-func (c context) deleteProductFiles() {
+func (c context) stopKeybase() {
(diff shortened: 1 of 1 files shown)


maxtaco:
Is there a way to solve this problem atomically?  On Linux, for instance, I'd move all install files into a temporary directory, and then move the directory into place to complete the install.  Can this work?

zanderz:
I don't think it would be worthwhile to try to get the windows installer mechanisms to do that - it is contrary to how they work. I think we'd be looking at a whole different kind of installer, maybe like Squirrel, or at least parts of it: https://medium.com/heresy-dev/auto-updating-apps-for-windows-and-osx-using-electron-the-complete-guide-4aa7a50b904c

PR #178, opened 2018-10-16
plumb through an ignore_snooze and use it in user initiated actions

songgao:
Related: https://github.com/keybase/keybase/pull/3022

### DIFF
--- a/keybase/context.go
+++ b/keybase/context.go
@@ -74,8 +74,8 @@ func newContextCheckCmd(cfg Config, log Log, isCheckCommand bool) *context {
 }
 
 // NewUpdaterContext returns an updater context for Keybase
-func NewUpdaterContext(appName string, pathToKeybase string, log Log, isCheck bool) (updater.Context, *updater.Updater) {
-	cfg, err := newConfig(appName, pathToKeybase, log)
+func NewUpdaterContext(appName string, pathToKeybase string, log Log, isCheck bool, ignoreSnooze bool) (updater.Context, *updater.Updater) {
+	cfg, err := newConfig(appName, pathToKeybase, log, ignoreSnooze)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


songgao:
bump!

keybase/context.go
```diff
@@ -74,8 +74,8 @@ func newContextCheckCmd(cfg Config, log Log, isCheckCommand bool) *context {
 }
 
 // NewUpdaterContext returns an updater context for Keybase
-func NewUpdaterContext(appName string, pathToKeybase string, log Log, isCheck bool) (updater.Context, *updater.Updater) {
-	cfg, err := newConfig(appName, pathToKeybase, log)
+func NewUpdaterContext(appName string, pathToKeybase string, log Log, isCheck bool, ignoreSnooze bool) (updater.Context, *updater.Updater) {
```
maxtaco: Not ideal to have a bunch of bools piling up on top of each other here. It's quite easy to mix them up. Do all 4 combinations make sense?

service/main.go
```diff
@@ -102,7 +102,7 @@ func run(f flags) {
 		svc.Run()
 	case "clean":
 		if runtime.GOOS == "windows" {
-			ctx, _ := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, true)
+			ctx, _ := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, true, false)
```
maxtaco: this sort of thing isn't great. `foo(true, false, true, true)`....

songgao:
Thanks @maxtaco; ready for another look!

songgao:
That’s one. Also people have complained that `keybase update check` doesn’t pull updates when it’s snoozed, so this should fix that too.

PR #181, opened 2018-12-14
Update deletes previous downloads

mlsteele:


maxtaco:
LGTM for macOS. @zanderz does this do the right thing on windows?

mlsteele:
Alright so @zanderz is this good to merge? Also this should be smoke tested. Is that part of the normal release process now?

zanderz:
yes, a release is always smoketested. I can run the build for you first if u don't have windows.

mlsteele:
That would be great

PR #186, opened 2019-12-05
Check if gui is busy after downloading the update

joshblum:
Not sure who is best to review here, feel free to un or re assign yoursef. Patch does the following:

- Checks if the gui is busy after downloading the app so we don't restart suddenly if the gui was not active before the download. There is still a chance the app restarts while the user is active but this should make the window much smaller. I've opened a ticket to think about having the gui enter an updating state so the user knows something is happening.
- Modies the ignoring gui busy check, to pass through after 10 hours instead of 3 attempts

cc @chrisnojima 

- [x] after release

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -312,7 +312,7 @@ type guiAppState struct {
 }
 
 func (u *Updater) checkUserActive(ctx Context) (bool, error) {
-	if time.Duration(u.guiBusyCount)*u.tickDuration >= time.Hour*10 { // Allow the update through after 10 hours
+	if time.Duration(u.guiBusyCount)*u.tickDuration >= time.Hour*6 { // Allow the update through after 6 hours
 		u.log.Warningf("Waited for GUI %d times - ignoring busy", u.guiBusyCount)
 		return false, nil
 	}
--- a/updater.go
+++ b/updater.go
@@ -307,7 +307,8 @@ func (u *Updater) promptForUpdateAction(ctx Context, update Update, options Upda
 }
 
 type guiAppState struct {
-	IsUserActive bool `json:"isUserActive"`
+	IsUserActive bool  `json:"isUserActive"`
+	ChangedAtMs  int64 `json:"changedAtMs"`
 }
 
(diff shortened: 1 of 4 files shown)


patrickxb:
I don't really know anything about the updater code, I think cjb is best.

updater.go
```diff
@@ -295,8 +311,7 @@ type guiAppState struct {
 }
 
 func (u *Updater) checkUserActive(ctx Context) (bool, error) {
-
-	if u.guiBusyCount >= 3 {
+	if time.Duration(u.guiBusyCount)*u.tickDuration >= time.Hour*10 { // Allow the update through after 10 hours
```
cjb: We used to wait max 3 hours, why the change to 10?

chrisnojima: I think we should leave it at 3

cjb: Yeah, something around 3 sounds better to me too.  It's nice to know that an update we push with urgency will go to everyone online within 4 hours, instead of within 11.

joshblum: seemed less likely to trigger during a normal work day when the app frequently being used

joshblum: 3 just means we will still all get app restarts throughout the day while using the app. the chance of needing an emergency update and the app being actively used for 11 hours seems slim, i strongly prefer leaving it on a longer window. the purpose of this check is really just in case we leave the busy file in an incorrect state so we eventually update, doesn't seem critical to have such a short time window.

cjb: Well, it means we'll get likely one app restart in the day.  That's probably not a bad thing, I don't want to spend all work day still running yesterday's admin build if there's a new one out.

joshblum: This is only if you're actively using the app during all of these checks. I find it much more frustrating to have the app swept out from under me.

joshblum:
depends on https://github.com/keybase/client/pull/21470

joshblum:
merging but not updating the version of the updater used for packaging yet (in case we hotfix in the near future) https://github.com/keybase/client/blob/master/packaging/desktop/package_darwin.sh#L97 cc @mmaxim

PR #187, opened 2020-01-02
more logging specifically for windows

xgess:


watchdog/watchdog.go
```diff
@@ -57,7 +57,7 @@ func terminateExisting(programs []Program, log Log) {
 		matcher := process.NewMatcher(program.Path, process.PathEqual, log)
 		matcher.ExceptPID(ospid)
 		log.Infof("Terminating %s", program.Path)
-		process.TerminateAll(matcher, time.Second, log)
+		process.TerminateAll(matcher, 3*time.Second, log)
```
patrickxb: i assume 3s is how long it waits?  any harm in making it 60s?

xgess: lol. actually, it looks like i was mistaken. this isn't even used in windows. but for mac and linux, it's how long we wait for something to die gracefully (SIGTERM) before it's slayed without prejudice (SIGKILL). and it does these fully in serial. which is probably something we should change. but either way, go and signals don't really work on windows. i think i should probably remove this change entirely.

PR #191, opened 2020-01-30
handle watchdog race condition

xgess:
windows has the unique distinction (i think) of being started by a process (the watchdog) which is not guaranteed by the operating system to be a singleton. as a result, there is an additional risk here that isn't present in the other desktop operating systems. 

if two watchdogs are started at the same time (or even within ~1sec of each other) it's possible to have multiple iterations of the service running at the same time, which causes a bad time. 

this change slows down the process of terminating existing programs to really solidify that the last watchdog started will win.

### DIFF
--- a/keybase/report_test.go
+++ b/keybase/report_test.go
@@ -58,7 +58,7 @@ func TestReportBadResponse(t *testing.T) {
 }
 
 func TestReportTimeout(t *testing.T) {
-	server := newServerWithDelay(updateJSONResponse, 5*time.Millisecond)
+	server := newServerWithDelay(updateJSONResponse, 100*time.Millisecond)
 	defer server.Close()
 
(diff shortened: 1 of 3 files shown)


xgess:
@zapu i think this is a better approach than https://github.com/keybase/go-updater/pull/190

keybase/report_test.go
```diff
@@ -58,7 +58,7 @@ func TestReportBadResponse(t *testing.T) {
 }
 
 func TestReportTimeout(t *testing.T) {
-	server := newServerWithDelay(updateJSONResponse, 5*time.Millisecond)
+	server := newServerWithDelay(updateJSONResponse, 100*time.Millisecond)
```
xgess: this is an unrelated test flake. CI runs slow, and this fails sometimes.

PR #193, opened 2020-03-03
If Etag match, also check digest if required

gabriel:
If there is an ETag match, we return before digest check. I don't think this is a security issue but we might as well check the digest if specified.

### DIFF
--- a/util/http.go
+++ b/util/http.go
@@ -185,6 +185,13 @@ func downloadURL(urlString string, destinationPath string, options DownloadURLOp
 	defer DiscardAndCloseBodyIgnoreError(resp)
 	if resp.StatusCode == http.StatusNotModified {
 		cached = true
+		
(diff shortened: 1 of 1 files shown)


maxtaco:
Hi! Thanks for the PR, can you fill me in on what the higher level goal is?  Is it that your client has been downloading updates for no good reason (repeatedly?) and not applying them? If so there might be another bug to investigate.

gabriel:
Oh I was thinking about making a generic version/fork of this updater I could use in another project, and while I was poking around, noticed this scenario with etags and digests. I didn't encounter any bugs or weirdness or anything.

It seems like doing the digest check even if the etag matches is the correct behavior, even if in the keybase scenario that code path is never used? Basically when I looked at it, it seemed like an oversight on my part and I should mention it.

PR #196, opened 2020-04-07
PICNIC-664

thebearjew:
This PR an extension of https://github.com/keybase/client/pull/21591 to the updater.

#### Intended Flow

1. The GUI will start a 1 hour loop which will make RPC calls to `GetUpdateInfo2`.
2. On macOS and Windows, `GetUpdateInfo2` will call `updater download-latest` to check for a download and cache it. On Linux, `GetUpdateInfo2` will make a request to the `pkg/check` on the API server.
3. `download-latest` on the updater will make a request to `pkg/update` to determine if client needs an update. If the client does, it will first look on disk via `FindDownloadedAsset` to see if we've already cached a keybase update with the same version. If a download does not exist on disk, it will download it from S3. In both cases, once an update exists on disk, it will verify the digest (sha-256) and signature (saltpack) of the asset. This is to maintain the integrity of our cache. If either the digest or signature fail, the asset will be removed from disk. If everything was successful, we print `true` to stdout.
4. If `updater download-lateset` returns true, then the GUI will update KeybaseFM to show that an update is available.
5. The user can then click `Install update`  from KeybaseFM. This will fire a different RPC that will call `updater apply-downloaded`.
6. `apply-downloaded` makes **another** request to `pkg/update` on the API server to ensure that the client still requires an update. If an update is needed then we look on disk for a matching asset. If no asset is found, we return false and attempt to remove any previously downloaded assets. If we did find a matching asset, again we verify the digest (sha-256) and signature (saltpack). If either of these two steps fail, we return false and remove any previously downloaded assets. At this point the digest and signature are valid, so we attempt to apply the update without checking if the GUI is busy because it almost certainly is (given the flow).
7. The only thing that could prevent applying the update immediately is if the user is currently uploading files to KBFS. At this point the updater will present a native prompt to the user indicating this issue. Note: this prompt is not new, it is/has always been a part of the update apply flow.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -45,6 +45,14 @@ type Update struct {
 	NeedUpdate  bool       `json:"needUpdate"`
 }
 
+func (u Update) missingAsset() bool {
+	if u.Asset == nil || u.Asset.URL == "" {
+		return true
+	}
+	return false
+
+}
+
 // UpdateOptions are options used to find an update
 type UpdateOptions struct {
 	// Version is the current version of the app
--- a/protocol.go
+++ b/protocol.go
@@ -46,10 +46,7 @@ type Update struct {
 }
 
 func (u Update) missingAsset() bool {
-	if u.Asset == nil || u.Asset.URL == "" {
-		return true
-	}
-	return false
+	return u.Asset == nil || u.Asset.URL == ""
 
 }
 
--- a/service/main.go
+++ b/service/main.go
@@ -91,6 +91,8 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+		// Keybase service expects to parse this output as a boolean.
+		// Do not change unless changing in both locations
 		fmt.Println(needUpdate)
 	case "check":
 		if err := updateCheckFromFlags(f, ulog); err != nil {
@@ -104,6 +106,8 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+		// Keybase service expects to parse this output as a boolean.
+		// Do not change unless changing in both locations
 		fmt.Println(updateAvailable && updateCached)
 	case "apply-downloaded":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
@@ -112,6 +116,8 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+		// Keybase service expects to parse this output as a boolean.
+		// Do not change unless changing in both locations
 		fmt.Println(applied)
 	case "service", "":
 		svc := serviceFromFlags(f, ulog)
--- a/service/main.go
+++ b/service/main.go
@@ -99,13 +99,12 @@ func run(f flags) {
 		}
 	case "download-latest":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.CheckPassive)
-		fmt.Println("JRY main.go download-latest", f)
-		availableAndDownloaded, err := updater.CheckAndDownload(ctx)
+		updateAvailable, updateCached, err := updater.CheckAndDownload(ctx)
 		if err != nil {
 			ulog.Error(err)
 			os.Exit(1)
 		}
-		fmt.Println(availableAndDownloaded)
+		fmt.Println(updateAvailable && updateCached)
 	case "apply-downloaded":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
 		if err := updater.ApplyDownloaded(ctx); err != nil {
--- a/service/main.go
+++ b/service/main.go
@@ -107,10 +107,12 @@ func run(f flags) {
 		fmt.Println(updateAvailable && updateCached)
 	case "apply-downloaded":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
-		if err := updater.ApplyDownloaded(ctx); err != nil {
+		applied, err := updater.ApplyDownloaded(ctx)
+		if err != nil {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+		fmt.Println(applied)
 	case "service", "":
 		svc := serviceFromFlags(f, ulog)
 		svc.Run()
--- a/service/main.go
+++ b/service/main.go
@@ -97,6 +97,21 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+	case "download-latest":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.CheckPassive)
+		fmt.Println("JRY main.go download-latest", f)
+		availableAndDownloaded, err := updater.CheckAndDownload(ctx)
+		if err != nil {
+			ulog.Error(err)
+			os.Exit(1)
+		}
+		fmt.Println(availableAndDownloaded)
+	case "apply-downloaded":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
+		if err := updater.ApplyDownloaded(ctx); err != nil {
+			ulog.Error(err)
+			os.Exit(1)
+		}
 	case "service", "":
 		svc := serviceFromFlags(f, ulog)
 		svc.Run()
--- a/service/main.go
+++ b/service/main.go
@@ -101,14 +101,14 @@ func run(f flags) {
 		}
 	case "download-latest":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.CheckPassive)
-		updateAvailable, updateCached, err := updater.CheckAndDownload(ctx)
+		updateAvailable, _, err := updater.CheckAndDownload(ctx)
 		if err != nil {
 			ulog.Error(err)
 			os.Exit(1)
 		}
 		// Keybase service expects to parse this output as a boolean.
 		// Do not change unless changing in both locations
-		fmt.Println(updateAvailable && updateCached)
+		fmt.Println(updateAvailable)
 	case "apply-downloaded":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
 		applied, err := updater.ApplyDownloaded(ctx)
--- a/service/main.go
+++ b/service/main.go
@@ -93,6 +93,7 @@ func run(f flags) {
 		}
 		// Keybase service expects to parse this output as a boolean.
 		// Do not change unless changing in both locations
+		// https: //github.com/keybase/client/blob/master/go/client/cmd_update.go
 		fmt.Println(needUpdate)
 	case "check":
 		if err := updateCheckFromFlags(f, ulog); err != nil {
@@ -108,6 +109,7 @@ func run(f flags) {
 		}
 		// Keybase service expects to parse this output as a boolean.
 		// Do not change unless changing in both locations
+		// https: //github.com/keybase/client/blob/master/go/client/cmd_update.go
 		fmt.Println(updateAvailable)
 	case "apply-downloaded":
 		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
@@ -116,8 +118,6 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
-		// Keybase service expects to parse this output as a boolean.
-		// Do not change unless changing in both locations
 		fmt.Println(applied)
 	case "service", "":
 		svc := serviceFromFlags(f, ulog)
--- a/updater.go
+++ b/updater.go
@@ -195,9 +195,6 @@ func (u *Updater) ApplyDownloaded(ctx Context) (bool, error) {
 
 	// Only report apply success/failure
 	applied, err := u.applyDownloaded(ctx, update, options)
-	if applied || err != nil {
-		report(ctx, err, update, options)
-	}
 	if err != nil && !applied {
 		report(ctx, err, update, options)
 	}
--- a/updater.go
+++ b/updater.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/keybase/go-updater/util"
@@ -184,6 +185,76 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
 	return update, nil
 }
 
+func (u *Updater) ApplyDownloaded(ctx Context) error {
+	options := ctx.UpdateOptions()
+
+	// 1. check with the api server again for the latest update to be sure that a new update has not come out since our last call to CheckAndDownload
+	u.log.Infof("Attempting to apply previously downloaded update")
+	update, err := u.checkForUpdate(ctx, options)
+
+	if err != nil {
+		report(ctx, findErr(err), update, options)
+		return findErr(err)
+	}
+
+	err = u.applyDownloaded(ctx, update, options)
+
+	if err != nil {
+		report(ctx, err, update, options)
+		return err
+	}
+	return nil
+
+}
+
+// ApplyDownloaded will look for an previously downloaded update and attempt to apply it without prompting.
+// CheckAndDownload must be called first so that we have a download asset avaiable to apply.
+func (u *Updater) applyDownloaded(ctx Context, update *Update, options UpdateOptions) error {
+	if update == nil || !update.NeedUpdate {
+		u.log.Infof("No previously downloaded update to apply since client is update to date")
+		return nil
+	}
+	u.log.Infof("Got update with version: %s", update.Version)
+
+	if update.Asset == nil || update.Asset.URL == "" {
+		u.log.Info("No update asset to apply")
+		return nil
+	}
+
+	// 2. check the disk via FindDownloadedAsset. Compare our API result to this result. If the downloaded update is stale, clear it and start over.
+	downloadedAssetPath, err := u.FindDownloadedAsset(update.Asset.Name)
+
+	if err != nil {
+		return err
+	}
+
+	if downloadedAssetPath == "" {
+		u.log.Infof("No downloaded asset found for version: %s", update.Version)
+		if err := u.CleanupPreviousUpdates(); err != nil {
+			u.log.Infof("Error cleaning up previous downloads: %v", err)
+		}
+		// Bail
+		return nil
+	}
+	update.Asset.LocalPath = downloadedAssetPath
+
+	// 3. otherwise use the update on disk and apply it.
+	// Make sure that at the end of this flow (success or failure, we actually remove the downloaded update)
+	defer u.CleanupPreviousUpdates()
+
+	u.log.Infof("Verify asset: %s", downloadedAssetPath)
+	if err := ctx.Verify(*update); err != nil {
+		return verifyErr(err)
+	}
+
+	tmpDir := os.TempDir()
+	if err := u.apply(ctx, *update, options, tmpDir); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (u *Updater) apply(ctx Context, update Update, options UpdateOptions, tmpDir string) error {
 	u.log.Info("Before apply")
 	if err := ctx.BeforeApply(update); err != nil {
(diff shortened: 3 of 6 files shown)


updater.go
```diff
@@ -184,6 +185,76 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+			u.log.Infof("Error cleaning up previous downloads: %v", err)
+		}
+		// Bail
+		return nil
+	}
+	update.Asset.LocalPath = downloadedAssetPath
+
+	// 3. otherwise use the update on disk and apply it.
+	// Make sure that at the end of this flow (success or failure, we actually remove the downloaded update)
+	defer u.CleanupPreviousUpdates()
```
joshblum, line 243: I think you can set this right after downloading completes successfully. if we hit `downloadedAssetPath == ""` we want to run this anyway. Should log the error though

updater.go
```diff
@@ -263,6 +334,62 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
...
+		return false, err
+	}
+
+	if !update.NeedUpdate {
+		return false, nil
+	}
+
+	// Linux updates don't have assets so it's ok to prompt for update above before
+	// we check for nil asset.
+	if update.Asset == nil || update.Asset.URL == "" {
```
joshblum, line 351: maybe this can be in a helper/method on `Update` since we use it in a few places now

updater.go
```diff
@@ -263,6 +334,62 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
 	return update.NeedUpdate, nil
 }
 
+func (u *Updater) CheckAndDownload(ctx Context) (availableAndDownloaded bool, err error) {
```
joshblum, line 337: maybe this can be split up into two bools, `available` and `downloaded` the RPC should only expect `available=true` but `downloaded` can be helpful for testing the cache behavior in tests

updater.go
```diff
@@ -263,6 +334,62 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
...
+			return false, downloadErr(err)
+		}
+		downloadedAssetPath = update.Asset.LocalPath
+		digestChecked = true
+	}
+	update.Asset.LocalPath = downloadedAssetPath
+
+	u.log.Infof("Verify asset: %s", downloadedAssetPath)
+	if err := ctx.Verify(*update); err != nil {
+		if tmpDir != "" {
```
joshblum, line 377: can stick this in a `defer func() { if tmpDir != "" && err != nil {...}}()`

updater.go
```diff
@@ -402,3 +530,68 @@ func (u *Updater) Cleanup(tmpDir string) {
 		}
 	}
 }
+
+// Inspect previously downloaded updates to avoid redownloading
+func (u *Updater) FindDownloadedAsset(assetName string) (matchingAssetPath string, err error) {
```
joshblum, line 535: note: add unit

joshblum:
nuke `service/service binary

updater.go
thebearjew: @joshblum I changed this so that the second return value is `updateCached` which is true if the update is already on disk **or** if it's just been downloaded.

joshblum: isn't this equivalent to `updateAvailable`?

service/main.go
```diff
@@ -97,6 +97,21 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+	case "download-latest":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.CheckPassive)
+		fmt.Println("JRY main.go download-latest", f)
+		availableAndDownloaded, err := updater.CheckAndDownload(ctx)
```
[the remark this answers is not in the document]
thebearjew: Added unit

service/main.go
```diff
@@ -97,6 +97,21 @@ func run(f flags) {
...
+		if err != nil {
+			ulog.Error(err)
+			os.Exit(1)
+		}
+		fmt.Println(availableAndDownloaded)
+	case "apply-downloaded":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
+		if err := updater.ApplyDownloaded(ctx); err != nil {
```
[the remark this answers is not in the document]
thebearjew: Added unit

updater.go
```diff
@@ -184,6 +185,76 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
 	return update, nil
 }
 
+func (u *Updater) ApplyDownloaded(ctx Context) error {
```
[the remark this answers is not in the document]
thebearjew: 👍 done

updater.go
thebearjew: Added unit

updater.go
```diff
@@ -184,6 +182,80 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+	}
+
+	// Only report apply success/failure
+	applied, err := u.applyDownloaded(ctx, update, options)
+	if applied || err != nil {
+		report(ctx, err, update, options)
+	}
+	if err != nil && !applied {
+		report(ctx, err, update, options)
+	}
```
heronhaye, line 203: second if condition seems covered by the first unless its meant to report twice?

thebearjew: whooop! forgot to remove the first conditional thanks.

service/main.go
```diff
@@ -97,6 +97,22 @@ func run(f flags) {
...
 			os.Exit(1)
 		}
+	case "download-latest":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.CheckPassive)
+		updateAvailable, updateCached, err := updater.CheckAndDownload(ctx)
+		if err != nil {
+			ulog.Error(err)
+			os.Exit(1)
+		}
+		fmt.Println(updateAvailable && updateCached)
```
joshblum, line 107: can you drop a comment here and below that the bool output is parsed by the service so we shouldn't change it unless both change

updater.go
```diff
@@ -115,10 +116,7 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
 	}
 	u.log.Infof("Got update with version: %s", update.Version)
 
-	// Linux updates don't have assets so it's ok to prompt for update above before
-	// we check for nil asset.
-	if update.Asset == nil || update.Asset.URL == "" {
-		u.log.Info("No update asset to apply")
+	if u.updateMissingAsset(update) {
```
joshblum, line 119: prefer `update.IsNil()`, callers can log if they want

updater.go
```diff
@@ -184,6 +182,77 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
 	return update, nil
 }
 
+func (u *Updater) ApplyDownloaded(ctx Context) (bool, error) {
+	options := ctx.UpdateOptions()
+
+	// 1. check with the api server again for the latest update to be sure that a new update has not come out since our last call to CheckAndDownload
```
joshblum, line 188: can you wrap text here?

updater.go
```diff
@@ -184,6 +182,77 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+	u.log.Infof("Attempting to apply previously downloaded update")
+	update, err := u.checkForUpdate(ctx, options)
+
```
joshblum, line 191: rm whitespace

updater.go
```diff
@@ -184,6 +182,77 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+	if err != nil {
+		return false, findErr(err)
+	}
+
+	// Only report apply success/failure
+	applied, err := u.applyDownloaded(ctx, update, options)
+	if err != nil && !applied {
```
joshblum, line 198: applied should never be true if there was an error, just combine the report in the error check below

updater.go
```diff
@@ -184,6 +182,77 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+		u.log.Infof("No previously downloaded update to apply since client is update to date")
+		return false, nil
+	}
+	u.log.Infof("Got update with version: %s", update.Version)
+
+	if u.updateMissingAsset(update) {
+		return false, nil
+	}
+
+	// 2. check the disk via FindDownloadedAsset. Compare our API result to this result. If the downloaded update is stale, clear it and start over.
```
joshblum, line 221: wrap

updater.go
```diff
@@ -184,6 +182,77 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+	downloadedAssetPath, err := u.FindDownloadedAsset(update.Asset.Name)
+	defer func() {
+		if err := u.CleanupPreviousUpdates(); err != nil {
+			u.log.Infof("Error cleaning up previous downloads: %v", err)
+		}
+	}()
+
```
joshblum, line 228: rm whitespace

updater.go
```diff
@@ -184,6 +182,77 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+	}
+	u.log.Infof("Got update with version: %s", update.Version)
+
+	if u.updateMissingAsset(update) {
+		return false, nil
+	}
+
+	// 2. check the disk via FindDownloadedAsset. Compare our API result to this result. If the downloaded update is stale, clear it and start over.
+	downloadedAssetPath, err := u.FindDownloadedAsset(update.Asset.Name)
+	defer func() {
```
joshblum, line 223: this should only happen on a successful download, can you move it below the error check?

updater.go
```diff
@@ -263,6 +332,65 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
...
+			if err := u.CleanupPreviousUpdates(); err != nil {
+				u.log.Infof("Error cleaning up previous downloads: %v", err)
+			}
+		}
+		if tmpDir != "" {
+			u.Cleanup(tmpDir)
+		}
+	}()
+	var digestChecked bool
+	updateAvailable = true
```
joshblum, line 363: nuke and just return true if no error

updater.go
```diff
@@ -263,6 +332,65 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
...
+		tmpDir = u.tempDir()
+		// This will set update.Asset.LocalPath
+		if err := u.downloadAsset(update.Asset, tmpDir, options); err != nil {
+			return false, false, downloadErr(err)
+		}
+		updateCached = true
+		digestChecked = true
+		downloadedAssetPath = update.Asset.LocalPath
+	}
+	updateCached = true
```
joshblum, line 376: i don't understand the purpose of this var anymore. it is equivalent to `updateAvailable`, as both will always be true if or false now. The original purpose was `updateDownloaded` since we wanted a flag for tests, why the change?

updater.go
```diff
@@ -365,7 +493,16 @@ func (u *Updater) tempDir() string {
 	return tmpDir
 }
 
+func (u *Updater) updateMissingAsset(update *Update) bool {
```
joshblum, line 496: move to member method on `Update`

updater.go
```diff
@@ -402,3 +539,59 @@ func (u *Updater) Cleanup(tmpDir string) {
...
 	}
 }
+
+// Inspect previously downloaded updates to avoid redownloading
+func (u *Updater) FindDownloadedAsset(assetName string) (matchingAssetPath string, err error) {
+	if assetName == "" {
+		return "", fmt.Errorf("No asset name provided")
+	}
+	parent := os.TempDir()
+
```
joshblum, line 549: rm whitespace

updater.go
```diff
@@ -402,3 +539,59 @@ func (u *Updater) Cleanup(tmpDir string) {
...
+	if parent == "" || parent == "." {
+		return matchingAssetPath, fmt.Errorf("temp directory is %v", parent)
+	}
+
+	files, err := ioutil.ReadDir(parent)
+
```
joshblum, line 555: rm whitespace

updater.go
```diff
@@ -402,3 +539,59 @@ func (u *Updater) Cleanup(tmpDir string) {
...
+	if err != nil {
+		return matchingAssetPath, fmt.Errorf("listing parent directory: %v", err)
+	}
+
+	stripParent := func(s string, p string) string {
```
joshblum, line 560: this is used once, no need for helper

updater.go
```diff
@@ -402,3 +539,59 @@ func (u *Updater) Cleanup(tmpDir string) {
...
+		return strings.TrimPrefix(s, p+string(filepath.Separator))
+	}
+
+	for _, fi := range files {
+		if !fi.IsDir() || !tempDirRE.MatchString(fi.Name()) {
+			continue
+		}
+
+		keybaseTempDirAbs := filepath.Join(parent, fi.Name())
+		walkErr := filepath.Walk(keybaseTempDirAbs, func(fullPath string, info os.FileInfo, inErr error) (outErr error) {
```
joshblum, line 570: `outErr` -> `err`

updater.go
```diff
@@ -402,3 +539,59 @@ func (u *Updater) Cleanup(tmpDir string) {
...
+
+			if info.IsDir() {
+				if fullPath == keybaseTempDirAbs {
+					return nil
+				}
+				return filepath.SkipDir
+			}
+
+			path := stripParent(fullPath, keybaseTempDirAbs)
+
```
joshblum, line 583: rm whitespace

updater_test.go
```diff
@@ -29,6 +29,26 @@ var testZipPath = filepath.Join(os.Getenv("GOPATH"), "src/github.com/keybase/go-
 
 var testAppStatePath = filepath.Join(os.TempDir(), "KBTest_app_state.json")
 
+// shasum -a 256 test/test.zip
+var validDigest = "54970995e4d02da631e0634162ef66e2663e0eee7d018e816ac48ed6f7811c84"
```
joshblum, line 33: make this `const`
```
const (
    validDigest = ".."
    ....
)

File: updater_test.go [hunk 27]
```diff
@@ -470,6 +490,253 @@ func TestUpdaterNotNeeded(t *testing.T) {
...
+	// FindDownloadedAsset = false
+	// return updateAvailable = false, updateCached = false
+	updateAvailable, updateCached, err := upr.CheckAndDownload(ctx)
+	assert.NoError(t, err)
+	assert.False(t, updateAvailable)
+	assert.False(t, updateCached)
+	assert.False(t, ctx.successReported)
+	assert.Equal(t, "deadbeef", upr.config.GetInstallID())
+
+	testUpdate.NeedUpdate = true
```
joshblum, line 517: put this under the comment about the case

updater_test.go
```diff
@@ -470,6 +490,253 @@ func TestUpdaterNotNeeded(t *testing.T) {
...
+	assert.True(t, updateCached)
+	assert.False(t, ctx.successReported)
+	assert.Equal(t, "deadbeef", upr.config.GetInstallID())
+
+	// 3.Find existing downloaded asset
+	// Need update = true
+	// FindDownloadedAsset = true
+	// return updateAvailable = true, updateCached = true
+	tmpDir := makeKeybaseUpdateTempDir(t, upr, testUpdate.Asset)
+
```
joshblum, line 533: rm whitespace

updater_test.go
```diff
@@ -470,6 +490,253 @@ func TestUpdaterNotNeeded(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.True(t, updateAvailable)
+	assert.True(t, updateCached)
+	assert.False(t, ctx.successReported)
+	assert.Equal(t, "deadbeef", upr.config.GetInstallID())
+
+	// 3.Find existing downloaded asset
+	// Need update = true
+	// FindDownloadedAsset = true
+	// return updateAvailable = true, updateCached = true
```
joshblum, line 531: we should check the the asset isn't re-downloaded in the case

updater.go
thebearjew: @joshblum update does not have `IsNil()`, is this something you want me to define?

updater.go
thebearjew: The real issue here was `return false, nil` when I should have just formatted an error.

updater.go
joshblum: yup!

updater.go
thebearjew: @joshblum `download-latest` should only return `true` to the service if an update is available via the API server and it's either:

1. Already on disk
2. We successfully downloaded it to disk.

Need both so that if a download fails, we don't return `true` to the service and therefore to the GUI. The GUI should only display an update if the download has been cached.

The original signature of this function had `availableAndDownloaded` as the named return value. It was split so that unit tests could check with more granularity.

updater_test.go
thebearjew: Good catch, thanks!

thebearjew:
Blocking this until release goes out. Our build script will pull master on `go-updater` so want to wait.

protocol.go
```diff
@@ -45,6 +45,14 @@ type Update struct {
 	NeedUpdate  bool       `json:"needUpdate"`
 }
 
+func (u Update) missingAsset() bool {
+	if u.Asset == nil || u.Asset.URL == "" {
```
joshblum, line 49: can be `return u.Asset == nil || u.Asset.URL == ""`

service/main.go
```diff
@@ -91,12 +91,34 @@ func run(f flags) {
 			ulog.Error(err)
 			os.Exit(1)
 		}
+		// Keybase service expects to parse this output as a boolean.
```
joshblum, line 94: can you add a permalink to the client file

service/main.go
```diff
@@ -91,12 +91,34 @@ func run(f flags) {
...
 			os.Exit(1)
 		}
+	case "download-latest":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.CheckPassive)
+		updateAvailable, _, err := updater.CheckAndDownload(ctx)
+		if err != nil {
+			ulog.Error(err)
+			os.Exit(1)
+		}
+		// Keybase service expects to parse this output as a boolean.
```
joshblum, line 109: permalink

service/main.go
```diff
@@ -91,12 +91,34 @@ func run(f flags) {
...
+	case "apply-downloaded":
+		ctx, updater := keybase.NewUpdaterContext(f.appName, f.pathToKeybase, ulog, keybase.Check)
+		applied, err := updater.ApplyDownloaded(ctx)
+		if err != nil {
+			ulog.Error(err)
+			os.Exit(1)
+		}
+		// Keybase service expects to parse this output as a boolean.
+		// Do not change unless changing in both locations
+		fmt.Println(applied)
```
joshblum, line 121: permalink

updater.go
```diff
@@ -184,6 +182,73 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+	u.log.Infof("Attempting to apply previously downloaded update")
+	update, err := u.checkForUpdate(ctx, options)
+	if err != nil {
+		return false, findErr(err)
+	}
+
+	// Only report apply success/failure
+	applied, err := u.applyDownloaded(ctx, update, options)
+	if err != nil {
+		report(ctx, err, update, options)
```
joshblum, line 199: `defer` instead

updater.go
```diff
@@ -184,6 +182,73 @@ func (u *Updater) update(ctx Context, options UpdateOptions) (*Update, error) {
...
+		return false, err
+	}
+	report(ctx, err, update, options)
+	return applied, nil
+
+}
+
+// ApplyDownloaded will look for an previously downloaded update and attempt to apply it without prompting.
+// CheckAndDownload must be called first so that we have a download asset avaiable to apply.
```
joshblum, line 208:
```suggestion
// CheckAndDownload must be called first so that we have a download asset available to apply.
```

updater.go
```diff
@@ -263,6 +328,63 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
...
 }
 
+func (u *Updater) CheckAndDownload(ctx Context) (updateAvailable, updateWasDownloaded bool, err error) {
+	options := ctx.UpdateOptions()
+	update, err := u.checkForUpdate(ctx, options)
+	if err != nil {
+		return false, false, err
+	}
+
+	if !update.NeedUpdate {
```
joshblum, line 338:
```suggestion
	if !update.NeedUpdate || update.missingAsset() {
```

updater.go
```diff
@@ -263,6 +328,63 @@ func (u *Updater) NeedUpdate(ctx Context) (upToDate bool, err error) {
...
+		return false, false, nil
+	}
+
+	if update.missingAsset() {
```
joshblum, line 342: can nuke this condition and combine with above

updater_test.go
```diff
@@ -29,6 +29,34 @@ var testZipPath = filepath.Join(os.Getenv("GOPATH"), "src/github.com/keybase/go-
 
 var testAppStatePath = filepath.Join(os.TempDir(), "KBTest_app_state.json")
 
+// shasum -a 256 test/test.zip
+const (
+	validDigest = "54970995e4d02da631e0634162ef66e2663e0eee7d018e816ac48ed6f7811c84"
+)
+
+// keybase sign -d -i test.zip
+const (
```
joshblum, line 38: slightly more readable to combine these into a single statement:
```
const (
  foo = "..."
  bar = "..."
)
```

PR #197, opened 2020-04-14
watchdog is safer about processes it kills when ExitAllOnSuccess

xgess:
* the biggest change here is around ExitAllOnSuccess. when a program that's ExitAllOnSuccess terminates gracefully, the watchdog will terminate all of its other programs and then itself. Previously, it looked for the other programs to kill by path. Now it keeps track of the current pid of each of its managed processes, and it precisely terminates those (and only those) pids. 
* there's a new test for the race condition ^ which fails on master and succeeds on this branch
* a bunch of changes for logging. the most noticeable will be that logs now display a human-readable `name` of the process. and there's a heartbeat; every hour, the watchdog will log all of the processes and pids that it thinks it's watching.

### DIFF
--- a/watchdog/watchdog.go
+++ b/watchdog/watchdog.go
@@ -24,11 +24,16 @@ const (
 	ExitAllOnSuccess ExitOn = "all"
 )
 
+const terminationDelay = 200 * time.Millisecond
+const heartbealDelay = 1 * time.Hour
+
 // Program is a program at path with arguments
 type Program struct {
-	Path   string
-	Args   []string
-	ExitOn ExitOn
+	Path       string
+	Name       string
+	Args       []string
+	ExitOn     ExitOn
+	runningPid int
 }
 
 // Log is the logging interface for the watchdog package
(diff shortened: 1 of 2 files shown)


watchdog/watchdog.go
```diff
@@ -39,95 +44,165 @@ type Log interface {
...
+		if p > 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func (w *Watchdog) killSimilarRunningPrograms() (killedPids []int) {
+	// kill any running processes that look like the ones this watchdog wants to watch
+	// this logic also exists in the updater, so if you want to change it, look there too.
```
taruti, line 198: This code needs to be changed in two places? Or is the comment old?

watchdog/watchdog.go
```diff
@@ -39,95 +44,165 @@ type Log interface {
...
+	ospid := os.Getpid()
+	for _, program := range w.Programs {
+		matcher := process.NewMatcher(program.Path, process.PathEqual, w.Log)
+		matcher.ExceptPID(ospid)
+		w.Log.Infof("Terminating %s", program.Name)
+		pids := process.TerminateAll(matcher, time.Second, w.Log)
+		killedPids = append(killedPids, pids...)
+	}
```
taruti, line 206: Perhaps logs pids and killedPids?

watchdog/watchdog.go
xgess: comment is still valid. the updater needs to shut down everything but itself before downloading and kicking off an update. it does that by (1) looking for the watchdog (i.e. the keybase service process which is the parent of another keybase service process) and killing it, then (2) executing similar code to this which kills everything that looks like a match except for its pid.

watchdog/watchdog.go
```diff
@@ -39,95 +44,165 @@ type Log interface {
...
-		log.Infof("Terminating any other programs we are monitoring")
-		terminateExisting(programs, log)
-		os.Exit(0)
+	w := Watchdog{
+		Programs:     programs,
+		RestartDelay: restartDelay,
+		Log:          log,
+		shutdownCh:   make(chan struct{}),
 	}
+	w.Log.Infof("Terminating any existing programs we will be monitoring")
```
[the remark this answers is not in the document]
xgess: that's almost exactly the line being logged three lines down

watchdog/watchdog.go
xgess: yeah - this gets logged by the caller of this function

watchdog/watchdog.go
```diff
@@ -39,95 +44,165 @@ type Log interface {
...
 }
 
-func watchPrograms(programs []Program, delay time.Duration, log Log, exitAll func()) {
-	for _, program := range programs {
-		go watchProgram(program, delay, log, exitAll)
+func (w *Watchdog) heartbeatToLog(delay time.Duration) {
+	// wait enough time for the first heartbeat so it's actually useful
+	time.Sleep(1 * time.Minute)
+	for {
+		var heartbeatData []struct {
```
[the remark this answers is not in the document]
xgess: sure

